PR #89, opened 2023-09-11
Attempt to fix background image/gradients

kafkas:


packages/ui/app/src/docs/BgImageGradient.tsx
```diff
@@ -0,0 +1,39 @@
...
+                    }
+                />,
+                document.body,
+                "bg-image-gradient"
+            )}
+        </>
+    );
+};
+
+export const BgImageGradient = dynamic(() => Promise.resolve(Core), {
```
dsinghvi, line 37: why are we making ssr false here?

packages/ui/app/src/docs/BgImageGradient.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { createPortal } from "react-dom";
+
+export declare namespace BgImageGradient {
+    export interface Props {
+        hasSpecifiedBackgroundColor: boolean;
+        hasSpecifiedBackgroundImage: boolean;
+    }
+}
+
+export const Core: React.FC<BgImageGradient.Props> = ({ hasSpecifiedBackgroundColor, hasSpecifiedBackgroundImage }) => {
```
dsinghvi, line 12: can we wrap this in a memo and export that component?

dsinghvi:
Screen Shot 2023-09-11 at 2 01 24 PM
not sure why dev failed

PR #94, opened 2023-09-11
Fix FontAwesome SSR error

kafkas:
Used [solution](https://github.com/FortAwesome/Font-Awesome/issues/19348#issuecomment-1262137893) suggested in the relevant issue.


dsinghvi:
@kafkas once we get a preview url with https://vellum-staging.docs.dev.buildwithfern.com/prompts/prompt-template-syntax and confirm it works, this is g2g

PR #98, opened 2023-09-12
Use `apiSection.showErrors` to conditionally display errors 

kafkas:


packages/ui/app/src/api-page/endpoints/EndpointContent.tsx
```diff
@@ -20,6 +21,7 @@ const URL_OVERFLOW_THRESHOLD = 0.95;
 
 export declare namespace EndpointContent {
     export interface Props {
+        apiSection: FernRegistryDocsRead.ApiSection;
```
dsinghvi, line 24: we should pass the minimal state

dsinghvi (approved):

PR #1288, opened 2024-08-10
Update ReadmeGenerator to link to Fern's Marketing Website

dannysheridan:

1. "SDK generated by Fern" to "Built with Fern"
2. URL from github.com/fern-api/fern to buildwithfern.com and a UTM link to track the organization's readme where this came from

Note: if we could track the language & organization, that'd be ideal. I'd like to understand that company A's Python SDK has 3x the inbound clicks to our website than their Ruby SDK. (In this suggested PR, there is no indication of which language SDK the user came from.)

We want prospective users to land on our marketing website, not our GitHub profile.

It has not been.


clis/generator-cli/src/readme/ReadmeGenerator.ts
```diff
@@ -218,7 +218,7 @@ export class ReadmeGenerator {
 
     private writeFernShield({ writer }: { writer: Writer }): void {
         writer.writeLine(
-            "[![fern shield](https://img.shields.io/badge/%F0%9F%8C%BF-SDK%20generated%20by%20Fern-brightgreen)](https://github.com/fern-api/fern)",
+            "[![fern shield](https://img.shields.io/badge/%F0%9F%8C%BF-Built%20with%20Fern-brightgreen)](https://buildwithfern.com?utm_source=github&utm_medium=readme&utm_campaign=${this.organizationPascalCase})",
```
amckinney, line 221:
```suggestion
            `[![fern shield](https://img.shields.io/badge/%F0%9F%8C%BF-Built%20with%20Fern-brightgreen)](https://buildwithfern.com?utm_source=github&utm_medium=readme&utm_campaign=${this.organizationPascalCase})`,
```

clis/generator-cli/src/readme/ReadmeGenerator.ts
```diff
@@ -217,8 +217,9 @@ export class ReadmeGenerator {
     }
 
     private writeFernShield({ writer }: { writer: Writer }): void {
+        const repoSource = this.readmeConfig.remote?.repoUrl ?? `${this.organizationPascalCase}/${this.languageTitle}`;
```
amckinney, line 220: The `repoUrl` will be in the form of `https://github.com/acme/acme-go` (i.e. it contains the `https://` prefix). Will that work if we include it in as the `utm_source`, or do we need to process this a little more?

Also, it might be best to just use `this.readmeConfig.organization` here (instead of the PascalCase-equivalent) since GitHub organizations often match the Fern organization name.

amckinney: Discussed offline - this is URL-encoded so we should be fine as long as this accurately tracks the source and makes it easy to inspect. 

We can always follow up with another lightweight change if this format isn't ideal.

amckinney (approved):

PR #1330, opened 2024-08-22
improvement(fern-bot): the bot stops using docker to upgrade generators

armandobelardo:
Fixes FER-2916 and FER-2232

This PR does a few things, the highlights:
- Uses the updated `fern generator upgrade` commands which no longer leverage docker, and so should run on Lambdas
- Splits the PRs per generator invocation
- Adds changelogs to the upgrade PRs

Example PRs:
https://github.com/fern-api/fern-bot-tests/pull/47
https://github.com/fern-api/fern-bot-tests/pull/45
https://github.com/fern-api/fern-bot-tests/pull/44

Biggest wackiness:
- We need to have a better way to pipe `yes` through the CLI, it seems non-trivial with `execa`, so right now we only pipe a single `y`, but in reality we might have several interactive steps and you really just want `yes | fern upgrade`, but this kept breaking due to the stdout effectively all buffering then erroring out when being piped to the fern command


servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
```diff
@@ -19,66 +21,261 @@ async function isOrganizationCanary(venusUrl: string, fullRepoPath: string): Pro
...
-    await getOrUpdateBranch(git, originDefaultBranch, GENERATOR_UPDATE_BRANCH);
-
     try {
         if (!(await isOrganizationCanary(venusUrl, fullRepoPath))) {
             console.log("Organization is not a fern-bot canary, skipping upgrade.");
             return;
         }
+    } catch (error) {
+        console.error("Could not determine if the repo owner was a fern-bot canary, quitting:", error);
+        return;
```
dsinghvi, line 152: can we add a sentry error here -- want to make sure we are 0ing out any exceptions here

servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
```diff
@@ -19,66 +21,261 @@ async function isOrganizationCanary(venusUrl: string, fullRepoPath: string): Pro
...
+    }
 
-        // Run fern CLI upgrade as well as generator upgrade which will go through each group in gen.yml
-        // and upgrade the generator version to the latest, non-RC version tagged in DockerHub.
-        await execFernCli("upgrade", fullRepoPath);
-        await execFernCli("generator upgrade", fullRepoPath);
+    try {
+        await updateSingleEntity({
```
dsinghvi, line 156: nit: what is an entity here? can we rename this to `handleSingleRepository` ? I think its a bit easier to understand

dsinghvi: ah just read more of the code here, so it handles an upgrade. `handleSingleUpgrade` ?

dsinghvi:
Looks great 🚀 

Once there's an ETE test that makes sure the action works for upgrading a CLI + upgrading a generator in a mock repo, then i think we are good to send it.

servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
armandobelardo: That's right, it's an upgrade of either the CLI or the generator, hence `entity`
I can update

servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
armandobelardo: added via their lambda layer, we'll see if it works

dsinghvi (approved):

PR #1374, opened 2024-09-03
improvement(fdr): break changelog endpoints to take in more specific version ranges

armandobelardo:
This PR breaks the existing versions endpoint to allow you to pass in exclusive or inclusive ends to your range, as opposed to just a from and to

The syntax of the input itself is kinda taken from openAPI

To allow fern bot to pass in the current versions within a repo as an exclusive min, and the new version as an inclusive max when grabbing the changelog for an upgrade

The ete tests below!


fern/apis/fdr/definition/generators/commons.yml
```diff
@@ -10,6 +10,13 @@ types:
           The remediation strategy to be used when a yank is detected. If one is not specified, we'll likely go backwards by publishing time
           to find the last unyanked version.
 
+  GetChangelogRequest:
+    properties:
+      from_version_inclusive: optional<string>
+      from_version_exclusive: optional<string>
+      to_version_exclusive: optional<string>
+      to_version_inclusive: optional<string>
```
amckinney, line 18: Might be worth modeling this so that it's impossible for `from_version_inclusive` to be set alongside `from_version_exclusive` (and same for `to_version_*`). Otherwise it's an input error, right?

I could imagine something like:

```yaml
GetChangelogRequest:
  properties:
    from_version: optional<VersionRange>
    to_version: optional<VersionRange>
```

Would this change be lightweight enough to include here?

armandobelardo: Presumably `VersionRange` is a union? I can do that, was just hoping to keep it simple since it's an internal API

dsinghvi: Similar thoughts, is this what you were imagining `VersionRange` would look like? 

```yml
GetChangelogRequest: 
  properties: 
    from_version: optional<Version> 
    to_version: optional<Version>

Version: 
  properties: 
    version: string
    inclusive: boolean
```

armandobelardo:
```
  VersionRange:
    union:
      inclusive: string
      exclusive: string

  GetChangelogRequest:
    properties:
      from_version: VersionRange
      to_version: VersionRange
```

armandobelardo: So many cooks in the kitchen! Thank goodness

amckinney (approved):
Awesome, very clean 👍

PR #1392, opened 2024-09-04
fix(fern-bot): update branch naming to not use colons for updates

armandobelardo:
Our update branch names used to include a `:` when updating a multi-api spec, but this isn't valid for git so updating the branches to not include that

Was: `fern/update/fern-typescript-express@fdr:local`
Now: `fern/update/fern-typescript-express@fdr/local`


dsinghvi (approved):
Would be nice to have a snapshot test on what the PR body ends up looking like so that its easy to see whats changing

PR #1400, opened 2024-09-05
fix: breadcrumbs point to the current page if overview page is configured

abvthecity:


dsinghvi (approved):
feels like this deserves a test

PR #1407, opened 2024-09-06
fix: init posthog and set the api host

abvthecity:
Posthog is still not firing on `uploadcare.com/docs`. This PR 
- enables debug mode in posthog
- switches to always use `NEXT_PUBLIC_` prefix for the posthog api key in the codebase
- sets the api_host config if it's changed.

### DIFF
--- a/packages/ui/docs-bundle/.env-cmdrc.cjs
+++ b/packages/ui/docs-bundle/.env-cmdrc.cjs
@@ -1,7 +1,7 @@
 module.exports = {
     "fern-dev": {
         NEXT_PUBLIC_FDR_ORIGIN: "https://registry-dev2.buildwithfern.com",
-        NEXT_PUBLIC_POSTHOG_API_KEY: "",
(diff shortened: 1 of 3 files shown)


packages/ui/docs-bundle/.env-cmdrc.cjs
```diff
@@ -1,7 +1,7 @@
 module.exports = {
     "fern-dev": {
         NEXT_PUBLIC_FDR_ORIGIN: "https://registry-dev2.buildwithfern.com",
-        NEXT_PUBLIC_POSTHOG_API_KEY: "",
+        NEXT_PUBLIC_POSTHOG_API_KEY: process.env.NEXT_PUBLIC_POSTHOG_API_KEY,
```
dsinghvi, line 4: i dont think we want this on dev

abvthecity: this is fine— i'm moving the environment secrets to vercel so that it's colocated with all the other environment secrets.
- disabled on preview deployments
- disabled on app-dev.buildwithfern.com
Screenshot 2024-09-06 at 12 22 30 PM

dsinghvi (approved):

PR #1421, opened 2024-09-09
fix(fern-bot): allow npm install to work on lambdas

armandobelardo:
Fixes FER-3203

This PR:
- updates the npm prefix to ensure we're writing to a writable dir in lambdas (previously only `npx` was overridden, this adds the additional config which fixes `npm install` which is needed by `fern upgrade`)
- reinstalls the fern cli to have it at a known location given the override above
- moves the fern org check for generator upgrades only, not CLI upgrades, this allows us to upgrade the fleet to a CLI version that is compatible with generator upgrades before we roll that out more widely

### DIFF
--- a/servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
+++ b/servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
@@ -140,20 +140,17 @@ export async function updateVersionInternal(
     console.log(`Cloning repo: ${repository.clone_url} to ${fullRepoPath}`);
     await cloneRepo(git, repository, octokit, fernBotLoginName, fernBotLoginId);
 
-    const organization = cleanStdout((await execFernCli("organization", fullRepoPath)).stdout);
-    console.log(`Found organization ID: ${organization}`);
-
-    try {
-        if (!(await isOrganizationCanary(organization, venusUrl))) {
-            console.log("Organization is not a fern-bot canary, skipping upgrade.");
-            return;
-        }
-    } catch (error) {
-        console.error("Could not determine if the repo owner was a fern-bot canary, quitting.");
-        throw error;
-    }
-
     const slackClient = new SlackService(slackToken, slackChannel);
+
+    let maybeOrganization: string | undefined;
+    try {
+        maybeOrganization = cleanStdout((await execFernCli("organization", fullRepoPath)).stdout);
+        console.log(`Found organization ID: ${maybeOrganization}`);
+    } catch (error) {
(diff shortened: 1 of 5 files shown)


servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
```diff
@@ -140,21 +140,18 @@ export async function updateVersionInternal(
...
-            return;
-        }
+        maybeOrganization = cleanStdout((await execFernCli("organization", fullRepoPath)).stdout);
+        console.log(`Found organization ID: ${maybeOrganization}`);
     } catch (error) {
-        console.error("Could not determine if the repo owner was a fern-bot canary, quitting.");
-        throw error;
+        console.error(
+            "Could not determine the repo owner, continuing to upgrade CLI, but will fail generator upgrades.",
+        );
```
dsinghvi, line 152: Are you eventually planning on tracking this state anywhere ? Or should we send a slack message here?

servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
```diff
@@ -174,9 +171,21 @@ export async function updateVersionInternal(
...
         },
         slackClient,
-        organization,
+        maybeOrganization,
     });
 
+    try {
+        if (maybeOrganization == null) {
+            throw new Error("No organization was found, quitting before generator upgrades.");
+        } else if (!(await isOrganizationCanary(maybeOrganization, venusUrl))) {
```
dsinghvi, line 180: I think venus is probably 500ing bc number of requests -- one option is to just to add an endpoint for all the orgs that are canaries and load it in one go so that you can reuse across lambda invocations

dsinghvi (approved):

servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
armandobelardo: It'd be nice to do in a different way since right now we're likely going to get this a ton due to CLI versions being low

So I'm thinking of not addressing immediately, and maybe tracking in along side the other metrics we have per repo

servers/fern-bot/src/functions/generator-updates/shared/updateGeneratorInternal.ts
armandobelardo: Good call, I'll do that in a follow on PR and see if it helps

PR #1426, opened 2024-09-09
feat: enable word wrapping in code block

abvthecity:
when adding `wordWrap` to the metastring, it will enable word wrapping in the code block.


dsinghvi (approved):
Can you add this to changelog.md ? And we can own adding to docs

PR #1448, opened 2024-09-11
feat: add display name to discriminated union variant

RohinBhargava:
Fixes FER-3238

- Adds display names and plumbs availability to discriminated union variants in API reference
- Updates FDR to accept displayNames from IR

- Cohere customer ask

TBD

- Maybe need to add to playground as well? Will investigate.


packages/ui/app/src/playground/form/PlaygroundDescriminatedUnionForm.tsx
```diff
@@ -63,7 +63,7 @@ export const PlaygroundDiscriminatedUnionForm = memo<PlaygroundDiscriminatedUnio
             discriminatedUnion.variants.map(
                 (variant): FernDropdown.Option => ({
                     type: "value",
-                    label: titleCase(variant.discriminantValue),
+                    label: titleCase(variant.displayName ?? variant.discriminantValue),
```
armandobelardo, line 66: mega nit, don't you wanna use the display name as is? e.g. not `titleCase`d?

Not sure how we treat other displaynames today

armandobelardo (approved):
Code itself makes sense, would need to see an example to make sure the name is used in all the right places

RohinBhargava:
> Code itself makes sense, would need to see an example to make sure the name is used in all the right places

Totally makes sense -- might need to release to FDR and then can test from local. Not sure how to do that without a merge

PR #1452, opened 2024-09-11
feat(fern-bot): begin tracking pull requests as well

armandobelardo:
Fixes FER-3256

This PR:
- begins tracking repos that fern-bot is installed on
- begins tracking PRs that fern-bot opens

The idea here is to:
- set up webhook listeners for installation of the app + creation of PRs to then track any opened PRs from fern-bot
    - we'll then ideally extend this into notifying us + customers on stale upgrade PRs
    - we'll then ideally extend this into notifying ourselves on check statuses


fern/apis/fdr/definition/generators/commons.yml
```diff
@@ -1,6 +1,9 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/fern-api/fern/main/fern.schema.json
 
 types:
+  OrganizationId:
+    type: string
```
dsinghvi, line 5: https://github.com/fern-api/fern-platform/blob/dcb0f740cdde5c34cec7398fdd85bf3034ab07cb/fern/apis/fdr/definition/commons.yml#L8-L10

fern/apis/fdr/definition/generators/git.yml
```diff
@@ -0,0 +1,194 @@
...
+
+imports:
+  commons: ./commons.yml
+
+types:
+  # Soon to add to our data model:
+  #   - Orgs (internal)
+  #   - Products (internal)
+  #   - Features (internal)
+  RepositoryId: string
```
dsinghvi, line 13: what is the format for this? is the idea this will be agnostic `github`, `gitlab`

armandobelardo: yeah that's right, I can make it a discriminated union, that's probably a good idea

fern/apis/fdr/definition/generators/commons.yml
armandobelardo: Updated!

dsinghvi:
@armandobelardo small thought on top level structure -- i think it could actually make sense to do something like `git.yml` directly at the top level instead of putting it in the `generators/` folder

fern/apis/fdr/definition/generators/git.yml
```diff
@@ -0,0 +1,201 @@
...
+      id: RepositoryId
+      name: string
+      owner: string
+      fullName:
+        type: string
+        docs: The full name of the repository, e.g. fern-api/fern. It includes the owner, as well as the name of the repository.
+      url: string
+      repositoryOwner:
+        type: fdrCommons.OrgId
+        docs: The Fern organization ID of the repository owner.
```
dsinghvi, line 33: we chatted about about a `info` field that tells us if this is a repo containing a fern config or a repo for the SDK

fern/apis/fdr/definition/generators/git.yml
```diff
@@ -0,0 +1,201 @@
...
+  fdrCommons: ../commons.yml
+
+types:
+  # Soon to add to our data model:
+  #   - Orgs (internal)
+  #   - Products (internal)
+  #   - Features (internal)
+  GithubRepositoryId:
+    properties:
+      id: string
```
dsinghvi, line 16: is the github repository id here a uuid ? if so, can we just document that so folks know what they are getting

fern/apis/fdr/definition/generators/git.yml
```diff
@@ -0,0 +1,201 @@
...
+        docs: The full name of the repository, e.g. fern-api/fern. It includes the owner, as well as the name of the repository.
+      url: string
+      repositoryOwner:
+        type: fdrCommons.OrgId
+        docs: The Fern organization ID of the repository owner.
+
+  GithubUser:
+    properties:
+      name: string
+      email: string
```
dsinghvi, line 38: technically a user can have multiple emails (not sure if we need to load all emails and match against them as opposed to primary)

fern/apis/fdr/definition/generators/git.yml
armandobelardo: It's an int64 for whatever reason
image


Wanted to stringify to keep it flexible

fern/apis/fdr/definition/generators/git.yml
```diff
@@ -0,0 +1,201 @@
...
+      request: PullRequest
+
+    deletePullRequest:
+      docs: Delete specified pull request.
+      method: DELETE
+      path: /pull-request/{id}
+      path-parameters:
+        id: string
+
+    getOrganizationRepositoryMap:
```
dsinghvi, line 168: thoughts on replacing this with a filter on the `listRepositories` endpoint?

dsinghvi (approved):

fern/apis/fdr/definition/git.yml
```diff
@@ -0,0 +1,238 @@
...
+            type: optional<string>
+            docs: |
+              The organization name of the repository owner to filter pull requests by (ex: fern-api).
+      response: ListRepositoriesResponse
+
+    upsertRepository:
+      docs: Update or create the specified repository.
+      method: PUT
+      path: /repository
+      request: FernRepository
```
dsinghvi, line 175: Do you have an example where we may need to update the repository? The only one I can imagine is if an existing repo changes the SDK language that its supposed to have in it 

(dont think this requires any changes to the PR, but want to understand how you are imagining a consumer would use this endpoint besides creating repos)

dsinghvi: Wouldn't we need a `deleteRepository` or something?

dsinghvi (approved):
API LGTM (didn't check the impl at all)

fern/apis/fdr/definition/git.yml
armandobelardo: I don't have an update example, but adding each endpoint and dao was a bit of a hassle so wanted to try to be a bit more comprehensive just in case we need it down the line

Will add a delete!

PR #1457, opened 2024-09-12
feat: add posthog event for broken link

chdeskur:
This PR adds a PostHog event to log when a broken link has been accessed.

### DIFF
--- a/packages/ui/docs-bundle/src/utils/getDocsPageProps.ts
+++ b/packages/ui/docs-bundle/src/utils/getDocsPageProps.ts
@@ -221,7 +221,7 @@ async function convertDocsToDocsPageProps({
         // see: https://nextjs.org/docs/app/api-reference/file-conventions/not-found
 
         try {
-            capturePosthogEvent("broken_link", {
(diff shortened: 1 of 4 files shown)


packages/ui/docs-bundle/src/utils/getDocsPageProps.ts
```diff
@@ -219,6 +220,15 @@ async function convertDocsToDocsPageProps({
         // however, we should consider rendering a custom 404 page in the future using the customer's branding.
         // see: https://nextjs.org/docs/app/api-reference/file-conventions/not-found
 
+        try {
+            capturePosthogEvent("broken_link", {
+                slug,
+            });
+        } catch (e) {
+            // eslint-disable-next-line no-console
+            console.error(e);
```
dsinghvi, line 229: i think we should take out this console error since even if the call fails, we probably dont want to make it a user facing error

chdeskur: roger that

PR #1458, opened 2024-09-12
fix(fern-bot): stop hardcoding python as the changelog candidate

armandobelardo:
This also updates the PR body to include the from version for more SA


servers/fern-bot/src/functions/generator-updates/shared/getGeneratorNameFromImageName.ts
```diff
@@ -0,0 +1,58 @@
...
+        case "fern-ruby-sdk":
+            return "ruby-sdk";
+        case "fern-ruby-model":
+            return "ruby-model";
+        // Misc.
+        case "fern-postman":
+            return "postman";
+        case "fern-openapi":
+            return "openapi";
+
```
dsinghvi, line 49: should this be a call to the API as opposed to hard coding here ? (load all generators, generator docker -> generator id?) That way we don't have to do this for php

armandobelardo: updating to use the getimagebyname endpoint

PR #1464, opened 2024-09-12
fix: paginated changelog

abvthecity:
Implements pagination on the long-scrolling changelog page

### DIFF
--- a/packages/ui/app/src/components/BottomNavigationNeighbors.tsx
+++ b/packages/ui/app/src/components/BottomNavigationNeighbors.tsx
@@ -1,0 +1,40 @@
+import { useAtomValue } from "jotai";
+import { FC, useMemo } from "react";
+import { NEIGHBORS_ATOM } from "../atoms/navigation";
+import { useToHref } from "../hooks/useHref";
+import { DocsContent } from "../resolver/DocsContent";
+import { BottomNavigationButtons } from "./BottomNavigationButtons";
+
+interface BottomNavigationNeighborsProps {
+    neighbors?: DocsContent.Neighbors;
+}
+
+export const BottomNavigationNeighbors: FC<BottomNavigationNeighborsProps> = (props) => {
+    const toHref = useToHref();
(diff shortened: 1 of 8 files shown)


packages/ui/app/src/components/BottomNavigationNeighbors.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { DocsContent } from "../resolver/DocsContent";
+import { BottomNavigationButtons } from "./BottomNavigationButtons";
+
+interface BottomNavigationNeighborsProps {
+    neighbors?: DocsContent.Neighbors;
+}
+
+export const BottomNavigationNeighbors: FC<BottomNavigationNeighborsProps> = (props) => {
+    const toHref = useToHref();
+    const neighborsFromAtom = useAtomValue(NEIGHBORS_ATOM);
```
RohinBhargava, line 14: does atom state refresh on new deployments?

abvthecity: the atom is client-side only so it's reset only on the client

RohinBhargava: what happens if there is a changelog entry and the boundary of each page changes? is it ok to have > 10 neighbors in some cases?

RohinBhargava: or am I misinterpreting what this is doing?

RohinBhargava (approved):

packages/ui/app/src/components/BottomNavigationNeighbors.tsx
abvthecity: Screenshot 2024-09-12 at 3 19 09 PM

abvthecity: it just renders the left-right buttons

PR #1477, opened 2024-09-13
fix: api playground loading state

abvthecity:
Fixes INC-98

- allows user to open the api playground even if `/resolve-api` hasn't finished loading
- renders a loading state for the api playground so that the user isn't confused
- breaks up the api playground endpoint form component to be more readable (with complementary skeletons)


abvthecity:
https://github.com/user-attachments/assets/64fd0d42-ff5b-4e82-931a-3427a1eb9f4f

RohinBhargava (approved):

PR #1489, opened 2024-09-16
chore(fdr): add tags and PR links to releases data model

armandobelardo:
Fixes FER-3248

This PR adds the concept of a tag to the release itself, as well as PR links to the changelog entry items

1. Wanting to enrich the changelog that we will eventually render within the docs
2. Wanting to clarify which changes impact which users (e.g. this release was only impacting the fern definition, etc)

Some db ete tests


fern/apis/fdr/definition/generators/cli.yml
```diff
@@ -6,19 +6,41 @@ imports:
...
+  CliReleaseTagEnum:
+    enum:
+      - name: fern_definition
+        value: "Fern Definition"
+      - name: openapi
+        value: "OpenAPI"
+      - name: asyncapi
+        value: "AsyncAPI"
+      - name: grpc
+        value: "gRPC"
```
dsinghvi, line 18: do you think there's any benefit of structuring here? my guess is string is fine and if we want to structure we just put all of that onus on the client (seed)

fern/apis/fdr/definition/generators/commons.yml
```diff
@@ -43,8 +43,13 @@ types:
     properties:
       type: ChangelogEntryType
       summary: string
+      pullRequestUrl:
+        type: optional<string>
+        docs: |
+          The URL to the pull request that introduced this change. This is optional for backcompat.
+          You do not always need to specify the URL as the CLI can fill it in with the current PR, but given you can choose to not
+          release within the same PR as the change is introduced, this should be specified for divorcing the two.
```
dsinghvi, line 51: https://github.com/palantir/conjure/blob/de3cc4992faeccdb8535d480203a7d25eff04e0f/changelog/4.10.1/pr-537.v2.yml#L5-L6

interesting how palantir does it, just posting as another data point

fern/apis/fdr/definition/generators/versions.yml
```diff
@@ -20,20 +19,19 @@ types:
       customConfigSchema:
         type: optional<string>
         docs: The JSON schema (stringified) for the custom config that this generator version supports.
+      tags:
+        type: optional<list<string>>
+        docs: Tags to categorize the Generator release.
```
dsinghvi, line 24: should we add PR url here as well?

dsinghvi (approved):

PR #1521, opened 2024-09-20
fix: highlighted json lines suspense is causing 500 on some pages

abvthecity:
https://github.com/fern-api/fern-platform/pull/1383 introduced a suspense-related bug that was causing the API reference to time out.

This PR fixes the issue by
1. refactoring the useHighlightLines file (renamed from getJsonHighlightLines)
2. use loadable atom to resolve the async call to `import("jsonpath")`
3. add tests for useHighlightLines
4. rips out getJsonHighlightLines from WebhookExample component (because it's handled by its child)


RohinBhargava (approved):

packages/ui/app/src/api-reference/examples/useHighlightJsonLines.ts
```diff
@@ -0,0 +1,139 @@
...
+ */
+export function getJsonLineNumbers(
+    jq: Awaited<typeof import("jsonpath")>,
+    json: unknown,
+    path: JsonPropertyPath,
+    start = 0,
+): HighlightLineResult[] {
+    const jsonString = JSON.stringify(json, undefined, INDENT_SPACES);
+
+    const part = path[0];
```
RohinBhargava, line 40: path itself can never be nullish, right?

abvthecity: no it cannot

PR #1532, opened 2024-09-23
chore: turn on noOptionalProperties and useBrandedStringAliases for fdr

abvthecity:
turns on noOptionalProperties and useBrandedStringAliases

---

**This is largely a noop change and is a prerequisite for**
1. introducing new navigation types and safely handling read vs write types in the api, which will let us add a new canonical url field: https://github.com/fern-api/fern-platform/pull/1526
2. plumbing through all new properties for API reference related to descriptions, defaults, min, max, regexp, etc. https://github.com/fern-api/fern-platform/pull/1523

---

**Limitions on the sdk generator that required some hackery**:`.fernignore` only ignores files that have been specifically edited, but will keep generated files around if no manual overrides exist. that means i couldn't ignore the `stream-wrappers` folder except by adding `//ignore` comment to each file.


servers/fdr/src/api/generated/.fernignore
```diff
@@ -0,0 +1,12 @@
...
+api/resources/api/resources/v1/resources/register/resources/*
+api/resources/api/resources/v1/resources/register/types/*
+api/resources/commons/types/*
+api/resources/docs/resources/v1/resources/write/types/*
+api/resources/docs/resources/v2/resources/write/types/*
+api/resources/generators/types/*
+api/resources/navigation/*
+api/resources/snippets/types/*
+api/resources/snippetsFactory/types/*
+api/resources/templates/types/*
```
dsinghvi, line 12: can you explain what we are modifying in the types here

dsinghvi (approved):

PR #1541, opened 2024-09-24
fix: cache endpoint definitions in `resolve-api` route

dsinghvi:
This PR introduces a caching layer to `APIDefinitionResolver`. 

The playground is way too slow to load. 

It has not really been tested, hoping to use staging links.

### DIFF
--- a/packages/ui/docs-bundle/src/server/DocsCache.ts
+++ b/packages/ui/docs-bundle/src/server/DocsCache.ts
@@ -37,15 +37,32 @@ export class DocsKVCache implements ApiDefinitionResolverCache {
         await kv.srem(`${PREFIX}:${this.domain}:visited-slugs`, ...slug);
     }
 
-    public async putResolvedEndpoint(endpoint: ResolvedEndpointDefinition): Promise<void> {
-        await kv.sadd(this.getResovledEndpointId(endpoint.id), endpoint);
+    async putResolvedEndpoint({
+        apiDefinitionId,
+        endpoint,
+    }: {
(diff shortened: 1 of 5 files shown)


packages/ui/fern-dashboard/src/lib/utils.ts
```diff
@@ -1,6 +0,0 @@
-import { type ClassValue, clsx } from "clsx"
-import { twMerge } from "tailwind-merge"
-
-export function cn(...inputs: ClassValue[]) {
```
abvthecity, line 4: Undo

packages/ui/docs-bundle/src/server/DocsCache.ts
```diff
@@ -35,4 +36,16 @@ export class DocsKVCache {
...
+    public async putResolvedEndpoint(endpoint: ResolvedEndpointDefinition): Promise<void> {
+        await kv.sadd(this.getResovledEndpointId(endpoint.id), endpoint);
+    }
+
+    public getResolvedEndpoint(id: APIV1Read.EndpointId): Promise<ResolvedEndpointDefinition | null | undefined> {
+        return kv.get(this.getResovledEndpointId(id));
+    }
+
+    private getResovledEndpointId(id: APIV1Read.EndpointId): string {
+        return `${PREFIX}:${this.domain}:resolved-endpoint:${id}`;
```
abvthecity, line 49: This needs to include the api definition id

abvthecity (changes requested):

PR #1543, opened 2024-09-25
chore: new docs and api fdr types

abvthecity:
This PR introduces a few changes to fdr-sdk:

## 1. frontmatter is in fdr-sdk

https://github.com/fern-api/fern-platform/blob/1d887360fecd14f5bb42ad6c54fb2fc3fa6f6e39/fern/apis/fdr/definition/docs/latest/frontmatter.yml#L10-L90

Eventually we can turn on serde so that we can validate frontmatter on the CLI-side.

## 2. MarkdownContent types is now in fdr-sdk

https://github.com/fern-api/fern-platform/blob/1d887360fecd14f5bb42ad6c54fb2fc3fa6f6e39/fern/apis/fdr/definition/docs/latest/__package__.yml#L7-L30

And `WithDescription` in the API Definition will begin to depend on it:

https://github.com/fern-api/fern-platform/blob/1d887360fecd14f5bb42ad6c54fb2fc3fa6f6e39/fern/apis/fdr/definition/api/latest/commons.yml#L12-L16

This lets us completely delete the following files:
- https://github.com/fern-api/fern-platform/blob/main/packages/ui/app/src/resolver/types.ts
- https://github.com/fern-api/fern-platform/blob/main/packages/ui/app/src/resolver/ApiDefinitionResolver.ts
- https://github.com/fern-api/fern-platform/blob/main/packages/ui/app/src/resolver/ApiEndpointResolver.ts
- https://github.com/fern-api/fern-platform/blob/main/packages/ui/app/src/resolver/ApiTypeResolver.ts

^ this is problematic right now because of dropped descriptions and metadata.

and instead, we just need to "visit" each description and resolve in-place.

## 3. a flattened api reference structure

```yaml
ApiDefinition:
  properties:
    id: rootCommons.ApiDefinitionId
    endpoints: map<rootCommons.EndpointId, endpoint.EndpointDefinition>
    websockets: map<rootCommons.WebSocketId, websocket.WebSocketChannel>
    webhooks: map<rootCommons.WebhookId, webhook.WebhookDefinition>
    types: map<rootCommons.TypeId, type.TypeDefinition>
    subpackages: map<commons.SubpackageId, commons.SubpackageMetadata>
    auth: optional<ApiAuth>
    globalHeaders: optional<list<type.ObjectProperty>>
```

The main difference between this structure and APIV1Read is that all endpoints, websockets, webhooks, etc are coalesced into a single root package, rather than require the frontend to visit all the subpackages.

## Next steps

This new structure makes pruning really easy and to do and will unlock CRUD-based operations on the API Definition:

```
- GET /api/fern-docs/api-definition/{apiDefinitionId}/endpoint/${endpointId}
- GET /api/fern-docs/api-definition/{apiDefinitionId}/websocket/${websocketId}
- GET /api/fern-docs/api-definition/{apiDefinitionId}/webhook/${webhookId}
```

- In the API Playground, instead of resolving the entire api definition, we can resolve one at a time and hold them in redux-like state.
- it becomes easier to tell when there's a cache miss in the frontend when the MarkdownContent is type string rather than object. This can let us implement state-while-revalidate in certain areas of the frontend for an extra performance gain.


fern/apis/fdr/definition/api/latest/namespace.yml
```diff
@@ -0,0 +1,13 @@
...
+  v1: ../v1/commons.yml
+
+types:
+  NamespaceId: v1.SubpackageId
+
+  Namespace:
+    properties:
+      id: NamespaceId
+      name: string
+      displayName: optional<string>
```
dsinghvi, line 13: what are namespaces/where are they used?

fern/apis/fdr/definition/api/latest/type.yml
```diff
@@ -0,0 +1,156 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/fern-api/fern/main/fern.schema.json
+
+imports:
+  commons: commons.yml
+  rootCommons: ../../commons.yml
+  read: ../v1/read/type.yml
+  v1: ../v1/commons.yml
```
dsinghvi, line 7: can we make this `v1Commons`

fern/apis/fdr/definition/api/latest/type.yml
```diff
@@ -0,0 +1,156 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/fern-api/fern/main/fern.schema.json
+
+imports:
+  commons: commons.yml
+  rootCommons: ../../commons.yml
+  read: ../v1/read/type.yml
```
dsinghvi, line 6: can we make this `v1Read`

dsinghvi: makes it clearer as you get ready to add v2, etc

fern/apis/fdr/definition/api/latest/websocket.yml
```diff
@@ -0,0 +1,53 @@
...
+  WebSocketMessageOrigin: v1.WebSocketMessageOrigin
+  ExampleWebSocketMessage: read.ExampleWebSocketMessage
+
+  WebSocketChannel:
+    extends:
+      - commons.WithDescription
+      - commons.WithAvailability
+    properties:
+      id: rootCommons.WebSocketId
+      authed: boolean
```
dsinghvi, line 22: thought on changing all of `authed` to be a `list<ApiAuth>` -- that way you can potentially have multiple auth schemes showed since we want to eventually make that change. (easier to do now since you are writing a migration)

dsinghvi (approved):

fern/apis/fdr/definition/api/latest/namespace.yml
abvthecity: whoops i need to delete this

PR #1544, opened 2024-09-25
feat(ui): cache endpoint pages and pairs

dsinghvi:
Because endpoint caching works, we now also cache pages and endpoint pairs. 

Try to get humanloop and workato API playground to load instantly. 

Testing on dev

### DIFF
--- a/packages/ui/app/src/resolver/ApiDefinitionResolver.ts
+++ b/packages/ui/app/src/resolver/ApiDefinitionResolver.ts
@@ -9,6 +9,7 @@ import type { FernSerializeMdxOptions } from "../mdx/types";
 import { ApiEndpointResolver } from "./ApiEndpointResolver";
 import { ApiTypeResolver } from "./ApiTypeResolver";
 import type {
+    ResolvedApiPageMetadata,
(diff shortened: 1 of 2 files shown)


packages/ui/app/src/resolver/ApiDefinitionResolver.ts
```diff
@@ -128,8 +133,8 @@ export class ApiDefinitionResolver {
                             );
                         }
                         const [nonStream, stream] = await Promise.all([
-                            this.definitionResolver.resolveEndpointDefinition(endpointPair.nonStream),
```
armandobelardo, line 131: Should we clean these functions out of `ApiEndpointResolver` so everything leverages the cache?

dsinghvi: There are a couple places we use this -- so i didn't want to inject the cache everywhere

dsinghvi: Wanted to focus on the `resolve-api.ts` route before expanding coverage

PR #1547, opened 2024-09-26
fix: canonical default version and noindex on hidden pages

abvthecity:
Versioned docs are still creating 2 duplicate canonical urls for each page on the default version because the default version gets copied without setting the canonical slug. This PR fixes that issue and updates the tests to reflect it. Also, hidden pages are excluded from indexing as well.

---

main test to understand this change:

https://github.com/fern-api/fern-platform/blob/7df08a87e92688fbdd209e5b55259bf00bfb96b6/packages/fdr-sdk/src/navigation/__test__/fixtures.test.ts#L63-L105


RohinBhargava (approved):

packages/fdr-sdk/src/navigation/NodeCollector.ts
```diff
@@ -58,19 +58,11 @@ export class NodeCollector {
         }
         FernNavigation.traverseNavigation(rootNode, (node, _index, parents) => {
             // if the node is the default version, make a copy of it and "prune" the version slug from all children nodes
-            const parent = parents[parents.length - 1];
```
RohinBhargava, line 61: The idea is that we do not need parent information to make the call if the version should be pruned or not?

RohinBhargava (approved):

packages/fdr-sdk/src/navigation/NodeCollector.ts
abvthecity: i read this part of the code over a few times and decided it wasn't really doing anything so i cleaned up the function a bit.

the main thing that got fixed was here: https://github.com/fern-api/fern-platform/pull/1547/files/7df08a87e92688fbdd209e5b55259bf00bfb96b6#diff-75f76f7a52c67fc8407aaf5d664e9ef7530c8f8364034f8062cb26c4bda15887R28

basically when "pruning" the default version's slugs we will assign the canonical slug to the original slug, so that it treats the versioned slug as the canonical one.

The effect is that a page like
https://humanloop.com/docs/getting-started/overview
is treated like it's an alias of
https://humanloop.com/docs/v5/getting-started/overview

where the latter url is the canonical url since it's the *real* slug.

the effect of this change, most importantly, is that google and bing will start to link to the canonical url (with the full v5) even though all the navigation within the docs will use the alias.

PR #1562, opened 2024-09-30
feat: editable playground environment

RohinBhargava:
Fixes FER-3328

- Allows environments to be edited in api reference and api playground; updates snippets and resolution for use in other methods.
- Introduces `PLAYGROUND_ENVIRONMENT_ATOM`, which allows for the storage of an environment URL that persists across loads.
- Basic validation on input url (make sure it is well formatted for parsing downstream logic).

- Customer asks

https://github.com/user-attachments/assets/c2a7ec9a-35d1-4cee-b9b8-6e9a42962d32


abvthecity:
@RohinBhargava nice! but there's a couple ux bugs from your video:
- why did double clicking app.sg.workato.com switch to www.workato.com?
- it was a bit odd that double clicking the environment ended up causing the basepath to be editable as well. i think the button should encapsulate the basepath so it's clear that both are editable
- the font size should be the same between the rendered environment (inside the button) and the text input.

PR #1552, opened 2024-09-27
chore: implement api-definition migrator and add fixtures

abvthecity:
## Motivation

This is a noop change and followup to https://github.com/fern-api/fern-platform/pull/1543. Additional context can be found here: https://github.com/fern-api/fern-platform/pull/1523. Another follow-up is needed to consume the new utilities from fdr-sdk inside of fern docs.

## New fdr-sdk utilities:

- [ApiDefinitionV1ToV2](https://github.com/fern-api/fern-platform/blob/6d369212a5fd7fdcbfec6c3ad8a7d6a050d9bb2b/packages/fdr-sdk/src/api-definition/migrators/v1ToV2.ts) ([test](https://github.com/fern-api/fern-platform/blob/ajiang/api-definition-v2/packages/fdr-sdk/src/__test__/fixtures.test.ts#L72-L74)) migrator + snapshot tests
- [ApiDefinition.unwrapReference, ApiDefinition.unwrapObjectType](https://github.com/fern-api/fern-platform/blob/6d369212a5fd7fdcbfec6c3ad8a7d6a050d9bb2b/packages/fdr-sdk/src/api-definition/unwrap.ts) ([test](https://github.com/fern-api/fern-platform/blob/6d369212a5fd7fdcbfec6c3ad8a7d6a050d9bb2b/packages/fdr-sdk/src/api-definition/__test__/unwrap.test.ts)) - this is an evolution of [@fern-ui/ui/src/resolver/types.ts](https://github.com/fern-api/fern-platform/blob/2fdd21cbb11ae95ecf469d14eaf993930abd6730/packages/ui/app/src/resolver/types.ts#L683-L742) which is used in the frontend to flatten type references and optionals. The main difference is that we now keep track of descriptions, availability, and default values while unwrapping (whereas perviously they were dropped).
- [ApiDefinition.coalesceAvailability](https://github.com/fern-api/fern-platform/blob/6d369212a5fd7fdcbfec6c3ad8a7d6a050d9bb2b/packages/fdr-sdk/src/api-definition/availability.ts) ([test](https://github.com/fern-api/fern-platform/blob/6d369212a5fd7fdcbfec6c3ad8a7d6a050d9bb2b/packages/fdr-sdk/src/api-definition/__test__/availability.test.ts)) - used by the unwrappers to determine the availability status of a property.
[...]

PR #1564, opened 2024-09-30
chore: migrated API Playground to using the V2 API Definition typings

abvthecity:
This PR is a follow-up to https://github.com/fern-api/fern-platform/pull/1552

Instead of calling `/api/fern-docs/resolve-api`, which resolves all endpoints, websockets, and webhooks across all the api definitions, this PR introduces 3 new getter functions:
- `GET /api/fern-docs/api-definition/{api}/endpoint/{endpointid}` ([link](https://github.com/fern-api/fern-platform/pull/1564/files#diff-33c258049988a70d0654de2f6a7aef507e1c509a15281bd85a752832d3751f68))
- `GET /api/fern-docs/api-definition/{api}/websocket/{websocketid}` ([link](https://github.com/fern-api/fern-platform/pull/1564/files#diff-484165c5fac8c4e5c15729af5a9eb155b10a324f4f2a29e5e4317bc54e7e770c))
- `GET /api/fern-docs/api-definition/{api}/webhook/{webhookid}` ([link](https://github.com/fern-api/fern-platform/pull/1564/files#diff-c474004b49fbdc74c393343134674d1096639e17a675f1d4769e032428b9beef))

Each of these endpoints:
1. loads the api definition from the the api.v1.read endpoint instead of docs.v2.read, which reduces the payload size transferred from FDR to NextJS.
2. prune the api definition to the specific endpoint that is being requested
3. resolve and cache all api definition descriptions using the new `Transformer` utility.

In the API Playground, 
- all references to `ResolvedApiDefinition` types have been removed in favor of the fdr-sdk types.
- endpoint definition + endpoint node + types are wrapped in a `EndpointContext`, which makes it easier to pass the context down to its children without specifying each of the parameters separately. This is useful because an endpoint node and endpoint definition is always paired together.
- useSWRImmutable is used to leverage SWR's cache
- the PlaygroundEndpointSelector is configured so that hovering over endpoints will trigger a prefetch, improving perceived speed.

---

### Caching

- ApiDefinitionLoader: [link](https://github.com/fern-api/fern-platform/blob/6a565da1d1cceb5b4d9c5a26c343366f9a40923d/packages/ui/docs-bundle/src/server/ApiDefinitionLoader.ts) does the bulk of the load -> prune -> cache work in the [load()](https://github.com/fern-api/fern-platform/blob/6a565da1d1cceb5b4d9c5a26c343366f9a40923d/packages/ui/docs-bundle/src/server/ApiDefinitionLoader.ts#L102-L150) method, and the [resolveDescriptions()](https://github.com/fern-api/fern-platform/blob/6a565da1d1cceb5b4d9c5a26c343366f9a40923d/packages/ui/docs-bundle/src/server/ApiDefinitionLoader.ts#L222-L268) method.

- ApiDefinitionCache: [link](https://github.com/fern-api/fern-platform/blob/6a565da1d1cceb5b4d9c5a26c343366f9a40923d/packages/ui/docs-bundle/src/server/ApiDefinitionCache.ts) shows how we leverage `mset` and `mget` to batch setting api definitions to parallelize and reduce server round-trips.

---
- [x] tested on flagright (ensure that application/json on multipart form is retained)


dsinghvi (approved):
Blind +1

packages/ui/app/src/playground/PlaygroundContext.tsx
```diff
@@ -1,34 +1,15 @@
...
     ssr: false,
 });
 
-const fetcher = async (url: string) => {
-    const res = await fetch(url);
-    return res.json();
-};
-
 export const PlaygroundContextProvider: FC = () => {
-    const key = useApiRoute("/api/fern-docs/resolve-api");
```
RohinBhargava, line 20: @abvthecity, can we bring this back?

PR #1569, opened 2024-10-01
feat(grpc-proxy): Refactor fern-bot as grpc-proxy server

dsinghvi:
Surge to refactor grpc proxy as a fargate server


packages/ui/fern-dashboard/src/lib/utils.ts
armandobelardo, line 1: this seems unrelated, could you add it back

dsinghvi:
was running into more issues, finally got the dumb lambda working

PR #1579, opened 2024-10-01
wip: introduce github webhook listener

armandobelardo:
Fixes FER-3335

This PR adds
- a github webhook listener, which we can build off of in the future (just add more `app.webhook.on`s)
- a listener for running checks, and creates a custom check on fern config repos that runs pre-configured scripts by-generator (e.g. `yarn install` & `yarn build`)

Will fast follow pushing a branch to the SDK repo with the updates so a user can see the changes, the PR was just getting complex enough that I wanted to keep it separate


servers/fern-bot/Dockerfile
armandobelardo, line 1: To avoid the limitless issues we'd face trying to install a bunch of language libraries in Lambda, we're trying to do that ahead of time within a docker container that the lambda will run on

PR #1580, opened 2024-10-01
feat(fdr): introduce scripts to generators so fern bot can build the SDKs

armandobelardo:
Title says it all, but following along on the RFC here: https://www.notion.so/buildwithfern/GitHub-SDK-Previews-b20e8201194446e490e4a7870128c8b8?pvs=4

This adds in the data model needed for fern-bot to be able to run checks on the generated SDK


fern/apis/fdr/definition/generators/__package__.yml
```diff
@@ -55,6 +61,20 @@ types:
         type: string
         docs: The name of the docker image to pull to run this generator.
 
+      # These scripts might be over specifiying, but also sets us up in a decent position to centralize information to generate CI actions files when the time comes
```
amckinney, line 64: Not needed now, but I wonder if we'll need to capture relationships between these scripts (i.e. "this depends on that").

In the current state, feels like it's safe to assume these things are linear, and the next always depends on the previous (which looks like what you had in mind).

fern/apis/fdr/definition/generators/__package__.yml
```diff
@@ -55,6 +61,20 @@ types:
...
+        docs: The steps necessary to prepare the environment to run any of the scripts below, for example installing yarn itself.
+      installScript:
+        type: optional<Script>
+        docs: The steps necessary for installing the generated resource, if any (ex - `yarn install`).
+      compileScript:
+        type: optional<Script>
+        docs: The steps for building the generated resource.
+      testScript:
+        type: optional<Script>
+        docs: The steps for running tests for the generated resource.
```
amckinney, line 76: Should we carve out a separate type for this? Something like a `ScriptConfig`? If so, we could also drop the `Script` suffix stutter on each of these property names.

amckinney (approved):

fern/apis/fdr/definition/generators/__package__.yml
armandobelardo: Yeah that's right, it's assumed to be linear right now

armandobelardo: Will put a todo to flag this as a possible need

fern/apis/fdr/definition/generators/__package__.yml
armandobelardo: Updated!

PR #1586, opened 2024-10-01
remove console statements from fern platform with nextjs

RohinBhargava:

- Adds config to remove console statements from compiled next js code

- Console statements sometimes leak into prod


abvthecity (changes requested):
I think this will cause all logs to disappear from production, including error logs.

PR #1592, opened 2024-10-03
fix: revert to prior behavior where hidden pages are indexable

abvthecity:
Reverts changes introduced in https://github.com/fern-api/fern-platform/pull/1547 where hidden pages were removed from indexing, and adds comments explaining why.

To clarify:
unlike the behavior set by other docs providers, hidden pages in fern docs ARE indexable, and DO show up in the sitemap. **Indexing must be explicitly disabled using `noindex=true` in the frontmatter.** This is so that fern docs makes fewer assumptions about how SEO is handled, and rely on the docs maintainer to be explicit.


dsinghvi:
@abvthecity im not sure hidden pages should be indexed by default; seems less intuitive + safer than not indexing and forcing customers to mark index as true

PR #1593, opened 2024-10-03
feat: tree-aware version switcher

abvthecity:
This PR enhances the previous implementation of the version switcher, where previously it only checks if the current node exists across versions. Now it will also ascend the tree and check if its parents exist across versions.

This allows the user to stay inside the same navigational "context" even if the current node doesn't exist across versions. For example:
- if an endpoint exists in v1, but not in v2, we will land on the first page of the api reference when switching versions
- if a section exists in v2 but not in v1, switching between versions will retain the same tab selection, but go to the overview page of that tab.

----

* algorithm for converting the `node` and `parents` into a ascending tree-list is unit tested
* version switching functionality is smoke-tested in PR preview


packages/ui/docs-bundle/src/server/__test__/withVersionSwitcherInfo.test.ts
```diff
@@ -0,0 +1,31 @@
...
+            { type: FernNavigation.VersionNode["type"]; id: string }
+        >({ type: "page", id: "a" }, [
+            { type: "version", id: "d" },
+            { type: "tab", id: "c" },
+            { type: "section", id: "b" },
+        ]);
+
+        expect(version).toEqual({ type: "version", id: "d" });
+        expect(nodes.map((node) => node.id)).toEqual(["a", "b", "c"]);
+    });
```
dsinghvi, line 17: this test is a little hard to grok, how does this work?

dsinghvi: oh nvm i get it now

packages/ui/docs-bundle/src/server/withVersionSwitcherInfo.ts
```diff
@@ -0,0 +1,154 @@
...
+ * In order to preserve the "context" of the current node when switching between versions, we need to check if
+ * the current node exists in the other versions. If it doesn't, we can traverse up the tree to find the closest
+ * page that exists on that version.
+ *
+ * For example, if the current node is `/docs/v1/foo/bar`, and the user switches to `/docs/v2`, we need to find
+ * if `/docs/v2/foo/bar` exists. If it doesn't, we can try `/docs/v2/foo` and so on.
+ *
+ * ASSUMPTIONS:
+ * This function relies on the slug and parent slugs to match between the versions.
+ * If the slug is overridden in frontmatter, there is not currently a way to determine the correct slug to switch to.
```
dsinghvi, line 38: how would we solve this edge case? (sounds like you would need to traverse nodes in the navigation?)

dsinghvi (approved):

PR #1596, opened 2024-10-03
feat: set basic jwt authentication allowlist

abvthecity:
### Pruning the navigation tree

The pages that do not require authentication are allowed to be cached, and thus they _do_ get served via SSG. In this case, the navigation tree needs to be pruned to show ONLY the pages that unauthenticated users are allowed to see.

This required adding the following: 
1. a BFS traverser where:
2. if the visited node should be deleted, it will be removed from its parent, and
3. if the node cannot be deleted from the parent, then it will be removed as well, and
4. if the parent node no longer contains any children (or content), also delete it (recursively)
5. finally, update all the `pointsTo` redirects after pruning the tree, so that redirects only point to public pages.

In other words: pruning applies to all leaf nodes (visitable in the browser). But, if any container (i.e. section, api reference, changelog, etc) is found to NOT contain children, it too will be deleted.

### Special handling for section overview pages:

If a section contains an overview page, but it is not included in the allowlist, then only the overview page itself will get deleted. But since section nodes are not leaf nodes and may contain children, we avoid deleting it unless it has no children left.

### Known edge case:

 if there's a stream/nonstream endpoint node (aka. EndpointPairNode), if only one of the two nodes are marked to be pruned, then both nodes will be pruned

### Applying auth on routes selectively

Adds the ability to set an `allowlist` and a `denylist` on pages and api routes using edge config, when the authentication scheme is set to `basic_token_verification`.

To be conservative, the denylist is prioritized over the allowlist, and anything that isn't captured in the allowlist will automatically be denied as well.

The allowlist/denylist leverages [path-to-regexp@6](https://github.com/pillarjs/path-to-regexp/tree/6.x) which allows wildcard and regex matching.

Some API routes must be explicitly allowed. for example:
- ~/sitemap.xml~
- ~/api/fern-docs/sitemap.xml (must be set on both the origin and rewrite)~
- ~/path/to/changelog.rss~
- ~/api/fern-docs/changelog (must be set on both the origin and rewrite)~
- ~/api/fern-docs/api-definition/[api]/endpoint/[endpoint] (**danger** if certain endpoints are enabled, the api routes also need to be explicitly enabled or else api playground will stop working).~
- **/api/fern-docs/search** (**danger**. this will enable search. if enabled, all records will be searchable from algolia publicly.)

**UPDATE (10/7)**: 82221667196f98b4df039eba198cb0cbaa5ee23c undos the strict auth checks applied to all of the api endpoints except for `/api/fern-docs/search`. Two reason:
1. sitemap and changelog can be pruned out of the box
2. api definitions require knowledge of the ApiDefinitionId and EndpointId, and FDR doesn't gate these requests upstream. We can update this to be more clever at gating specific endpoints in the future by utilizing the pruner; but we'd need to cache the navigation nodes otherwise this endpoint will become extraordinarily slow.

### What's next

- there should be a feature flag that, when set, instead of outright deleting all the authenticated pages, it should mark them as "locked" but still visible in the frontend.
- enable audience filters using the pruning mechanism
- enable launchdarkly feature flags using the pruning mechanism
- move the custom auth config (allowlist) into docs.yml
- segment algolia search records between authenticated/unauthenticated, and other audiences.

## Tests

- Repo: https://github.com/fern-api/test-jwt-auth
- Deployment: https://test-jwt-auth-smoky.docs.buildwithfern.com/
- Configuration:
```
"allowlist": [
  "/",
  "/my-section/public-page",
  "/api-reference/imdb/get-movie"
]
```

Screenshot 2024-10-04 at 5 09 45 PM


dsinghvi (approved):
I think that we will need to productize this pretty fast (i.e. support audiences in docs.yml, multiple search indexes with the upcoming docs deployments)

PR #1597, opened 2024-10-04
feat(docs): update callouts to include icon attribute

chdeskur:
This PR adds the `icon` attribute to the Callout options so that custom icons can be used in the various Callout-styled components.  

The example below tests the PR using a custom icon with the <Info> components

Screenshot 2024-10-04 at 9 55 33 AM

### DIFF
--- a/packages/ui/app/src/mdx/components/callout/Callout.tsx
+++ b/packages/ui/app/src/mdx/components/callout/Callout.tsx
@@ -101,53 +101,41 @@ export const Callout: FC<PropsWithChildren<Callout.Props>> = ({ intent: intentRa
 
 // aliases
 
-export function InfoCallout({ children, title, icon }: PropsWithChildren<Omit<Callout.Props, "intent">>): ReactElement {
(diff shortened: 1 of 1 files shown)


abvthecity (approved):

packages/ui/app/src/mdx/components/callout/Callout.tsx
```diff
@@ -101,65 +101,85 @@ export const Callout: FC<PropsWithChildren<Callout.Props>> = ({ intent: intentRa
 
 // aliases
 
-export function InfoCallout({ children, title }: PropsWithChildren<Omit<Callout.Props, "intent">>): ReactElement {
+export function InfoCallout({ children, title, icon }: PropsWithChildren<Omit<Callout.Props, "intent">>): ReactElement {
```
abvthecity, line 104: you can also just spread the props in case new props are ever added in the future

change `{ children, title, icon }` to `{ children, ...props }`

```
<Callout intent="info" {...props}>
  {children}
</Callout>
```

PR #1608, opened 2024-10-08
chore: API Reference now uses Api Definition V2

abvthecity:
In this PR:
- delete Resolved API types
- delete the `/api/fern-docs/resolve-api` endpoint
- change all `shape: TypeReference` to `shape: TypeShape` because TypeShape is more flexible (allows you to specify TypeReference via an `alias`. This lets us unwrap websocket variants more smoothly.
- deletes schemaWithExample
- updates the resolveDocsContent to adopt the new typings
- update all api reference components to use the new typings
- in some cases, fixes the `descriptions` field (but not in all cases)
- fixes dropped globalHeaders in api playground
- delete the InternalTypeDefinitionError (in favor of just InternalTypeDefinition) for simplicity (also it's not implemented correctly anyway)
- refactor ApiPackageContents -> ApiDefinitionContent to use a flattened array instead of recursion. (eventually this will be used to collapse / summarize api sections)
- improved memoization of ApiDefinition
- update tests and snapshots

the following functionality has been commented out because it no longer works:
- EndpointRequestSnippet no longer links back to the api reference (we can precompute this somewhere)


fern/apis/fdr/definition/api/latest/type.yml
```diff
@@ -60,7 +60,7 @@ types:
       - commons.WithAvailability
     properties:
       key: rootCommons.PropertyKey
-      valueShape: TypeReference
```
dsinghvi, line 63: could this break any consumer?

abvthecity: no it won't, all the consumers use `unwrapReference` which accepts `TypeShape | TypeReference`

dsinghvi (approved):

dsinghvi:
didn't mean to auto merge this cause it would be valuable to understand the `stringHasMarkdown` change

PR #1625, opened 2024-10-09
fix: inconsistent deserialization of authprops

abvthecity:
Note: authProps.token always has a prefix `(like custom_)`


packages/ui/docs-bundle/src/server/withInitialProps.ts
```diff
@@ -61,9 +61,9 @@ export async function withInitialProps({
     const featureFlags = await getFeatureFlags(xFernHost);
 
     const authConfig = await getAuthEdgeConfig(xFernHost);
-    const loader = DocsLoader.for(xFernHost, auth?.token)
```
abvthecity, line 64: this is the crux of the issue. `auth.token` has prefix `custom_` and cannot be deserialized)

RohinBhargava (approved):

PR #1628, opened 2024-10-10
fix: actually get extra-properties wired up in fdr

dsinghvi:


packages/fdr-sdk/src/converters/db/convertAPIDefinitionToDb.ts
```diff
@@ -644,6 +645,7 @@ function transformDiscriminatedVariant({
             properties: writeShape.additionalProperties.properties.map((property) =>
                 transformProperty({ writeShape: property }),
             ),
+            extraProperties: undefined,
```
abvthecity, line 648: Missed one

abvthecity (changes requested):

packages/fdr-sdk/src/converters/db/convertAPIDefinitionToDb.ts
dsinghvi: dont think we need it in discriminated union variants

dsinghvi: but fixed as well

PR #1636, opened 2024-10-10
chore(fern-bot): update stale notifs to still update dev fern orgs

armandobelardo:


dannysheridan (approved):

servers/fern-bot/src/functions/stale-notifs/actions/sendStaleNotifications.ts
```diff
@@ -31,7 +31,7 @@ export async function sendStaleNotificationsInternal(env: Env): Promise<void> {
     const orgPullMap = new Map<string, PullRequest[]>();
     let staleBotPRsFound = false;
     for await (const pull of botPulls) {
-        if (EXCLUDE_ORGS.has(pull.repositoryOwner)) {
+        if (env.ENVIRONMENT !== "development" && EXCLUDE_ORGS.has(pull.repositoryOwner)) {
```
amckinney, line 34: Should we also check if `ENVIRONMENT` is not empty, or is empty "production" in this case?

amckinney (approved):

servers/fern-bot/src/functions/stale-notifs/actions/sendStaleNotifications.ts
armandobelardo: it should never actually be empty, so feel decent about this
We assert it isn't in `env.ts`

PR #1640, opened 2024-10-10
proposal: new edge config schema

abvthecity:
Today, our edge config looks like this:
```yaml
feature-flag-1:
 - customer-1
 - customer-2
feature-flag-2:
 - customer-1
authentication:
 customer-1.com:
   a: b
   c: d
 customer-1.docs.buildwithfern.com:
   a: b
   c: d
```

This proposal will change the config to be:
```yaml
customer-1.com:
  feature-flags:
    feature-flag-1: true
    feature-flag-2: true
  authentication:
    a: b
    c: d
customer-2.com:
  feature-flags:
    feature-flag-1: true
# config aliases:
customer-1.docs.buildwithfern.com: customer-1.com
customer-2.docs.buildwithfern.com: customer-2.com
```

----

### Why

This is a better data model because:
`get(host)` gets only the relevant config for that customer, and not all the customers. (shinks the edge config data transfer)

---

### Why not

preview deployments are not captured in this model

### DIFF
--- a/packages/ui/fern-docs-edge-config/package.json
+++ b/packages/ui/fern-docs-edge-config/package.json
@@ -37,6 +37,7 @@
     "@fern-ui/fern-docs-auth": "workspace:*",
     "@fern-ui/fern-docs-utils": "workspace:*",
     "@vercel/edge-config": "^1.1.0",
+    "ts-essentials": "^10.0.1",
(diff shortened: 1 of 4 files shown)


dsinghvi (approved):
yes please

PR #1654, opened 2024-10-11
[do not merge] fix: support different query parameter encoding

dsinghvi:
Fixes FER-


fern/apis/fdr/definition/api/v1/read/endpoint.yml
```diff
@@ -65,6 +65,12 @@ types:
     properties:
       key: string
       type: type.TypeReference
+      arrayEncoding: optional<QueryParameterArrayEncoding>
```
abvthecity, line 68: mark this as default: comma?

fern/apis/fdr/definition/api/latest/endpoint.yml
```diff
@@ -37,6 +37,21 @@ types:
...
+      commaSeparated:
+        type: type.ObjectProperty
+        docs: An array of query parameters that are serialized as a comma delimited list
+      exploded:
+        type: type.ObjectProperty
+        docs: An array of query parameters that are sent as individual params in the URL
+      deepObject:
+        type: type.ObjectProperty
+        docs: A query parameter that is an object with properties and serialized as `a[key1]=value1`
+      basic:
```
abvthecity, line 51: should we
- fall back to comma-separated if basic is an array?
- fall back to deep object if basic is an object, or an array with objects?

abvthecity (approved):

PR #1655, opened 2024-10-11
Revert "fix: added dismissable layer and upgraded radix packages"

RohinBhargava:
Reverts fern-api/fern-platform#1595


packages/ui/components/src/FernScrollArea.scss
```diff
@@ -21,7 +21,6 @@
 
         &[data-scrollbars="vertical"] > div {
             max-width: 100%;
-            width: 100%;
```
RohinBhargava, line 24: @abvthecity this is the change I had to make for the dropdown. I believe we can do the same for the scroll area

abvthecity (approved):

PR #1667, opened 2024-10-13
fix: ete workflow call

abvthecity:
**BLUF**: I've given up on debugging this, and introduced a hack for 404: https://github.com/fern-api/fern-platform/pull/1667/files#diff-6684b5aa3d32d957a995c4aa25302b49795827ad66205a768db1633404e174caR37

----

### The long winded reason:

nextjs pages routing is pretty screwed up with middleware. Honestly this alone is a pretty good reason to switch to app router (not to mention the [huge perf gains](https://workos.com/blog/migrating-to-next-js-app-router-with-zero-downtime#using-the-app-router-an-rsc-react-query-hybrid)).

The middleware attempts to do some url parsing magic that doesn't jive with the reverse proxy setup that we have setup. Meaning, when you rewrite `/subpath/_next/data/xxx/page.json` to `/_next/data/xxx/static/host/page.json`, nextjs thinks you've rewritten to `/_next/data/xxx/_next/data/xxx/static/host/page.json.json`.

As a result, instead of returning the data props as a json, nextjs exits and returns the HTML doc of the 404 page.

So when you rewrite `/subpath/_next/data/xxx/404.json` -> `/_next/data/xxx/404.json` it will rewrite to `/_next/data/xxx/_next/data/xxx/404.json.json` and then return an HTML page instead.

When the client router detects that server didn't return a JSON, it will ignore the hardNavigate404=true flag and do a client-side navigation to the 404 page, but when the client router DOES receive a JSON, it will actually do a hard navigate to the 404 page. (i gave up on finding out why)

So in short, I tried all of the permutations of the rewrites + nextjs settings, and found that:
- rewriting to `/_next/data/xxx/page.json` works in _development_ but does NOT work in _production_
- rewriting to `/page` + `x-nextjs-data=1` does work in _production_, but NOT in _development_. 
- rewriting to `/_next/data/xxx/404.json` works in _development_, but does NOT work in _production_. 
- rewriting to `/404` in does NOT work in _production_ either. (it's impossible)

To workaround this, in the middleware we'll always return `{}` when `/_next/data/xxx/404.json` is detected in the route. This is fine for now b/c our `/404.tsx` doesn't implement any SSG props. In the future, if we customize the 404 page to use SSG, this will break (added inline comment to warn)

-----

### Why migrate to the App Router:

We won't need to maintain a reverse proxy anymore.

Most of the pains here relate to the fact that pages router constrains our ISR caching to be keyed on the router path. This is why we have the `/static/[domain]/[[...slug]]` pattern, and a forward proxy implementation. Additionally, for pages that require cookies, we have to turn off ISR, i.e. `/dynamic/[domain]/[[...slug]]`, which completely negates the benefits of ISR.

The app router won't have these constraints, because all you need is a catch-all route:
`/[[...slug]]`

and ISR is not needed anymore; instead, the app router will let us make more granular caching decisions anywhere in the project. (authed users will notice a major speed up).

And `/subpath/_next/data/xxx/page.json` rewrites are no longer needed, because all the data is streamed in the following way: `/[[...slug]]?_rsc=1`.

Additionally, app router lets you customize the 404 page with server-side rendering + caching, which is a huge plus.

-----

### Other changes in this PR

- added another playwright spec for nextjs-to-nextjs forward proxy (mirrors the test we have for express-to-nextjs forward proxy)
- reorganized the playwright tests slightly to be flatter
- use explicit string literals for rewriting paths, because `NextUrl.clone()` actually bakes in some [implicit behavior around serializing and deserializing `/_next/data/` routes](https://github.com/vercel/next.js/blob/canary/packages/next/src/server/web/next-url.ts#L79).
- switches raw `exec` to `loggingExeca` in the vercel-scripts so that we have better logging in github actions.
- fixes how secrets are passed into the workflow_call for the playwright.yml github workflow


packages/ui/docs-bundle/turbo.json
```diff
@@ -3,7 +3,8 @@
   "extends": ["//"],
   "tasks": {
     "docs:build": {
-      "outputs": [".next/**"],
+      "outputs": [".next/**", "!.next/cache/**"],
```
abvthecity, line 6: this improves build time slightly

packages/ui/docs-bundle/src/server/xfernhost/edge.ts
```diff
@@ -22,7 +22,9 @@ export function getDocsDomainEdge(req: NextRequest, useSearchParams = false): st
         }
     }
 
-    throw new Error("Could not determine xFernHost from request.");
+    // eslint-disable-next-line no-console
+    console.error("Could not determine xFernHost from request. Returning buildwithfern.com.");
+    return "buildwithfern.com";
```
abvthecity, line 27: i kept on hitting 500 errors unexpectedly because we were throwing here. so this is added as a convenience.

packages/ui/docs-bundle/src/middleware.ts
```diff
@@ -28,15 +32,17 @@ export const middleware: NextMiddleware = async (request) => {
     if (pathname === "/404" || pathname === "/500" || pathname === "/_error") {
         const headers = new Headers(request.headers);
 
-        if (request.headers.get("referer")?.includes("/_next/data/")) {
-            headers.set("x-nextjs-data", "1");
+        if (request.nextUrl.pathname.includes("/_next/data/") && pathname === "/404") {
+            // This is a hack to mock the 404 data page, since nextjs isn't playing nice with our middleware
+            return NextResponse.json({}, { status: 404 });
```
abvthecity, line 37: this is the hack

packages/ui/docs-bundle/next.config.mjs
```diff
@@ -73,6 +73,8 @@ const nextConfig = {
         externalMiddlewareRewritesResolve: true,
     },
 
+    skipMiddlewareUrlNormalize: true,
```
abvthecity, line 76: we need this to be true otherwise nextjs will normalize `/_next/data/xxx/404.json` into `/404`. We only want to return `{}` when matching on the next-data route.

dsinghvi (approved):

PR #1683, opened 2024-10-21
feat: sanitize acorns and JSX components

abvthecity:
https://www.notion.so/buildwithfern/Escaping-mdx-expressions-126b23082e4c802fad95c64ed2b7d5a8

### Motivation
- unblock creating mdx-component imports via `mdx-bundler` + `@mdx-js/esbuild`
- reduce the amount of frontend errors thrown due to unescaped `{` and `<`

Why: to improve how the docs is rendered

---

### Supporting esm imports of mdx files

The reason why `import Component from "./path/to/snippet.mdx"` fails is because of `rehype-layout`, which wraps each markdown file with a mdxJsxFlowElement that denotes the layout, i.e. `GuideLayout`. This was done because it was convenient to parse the table of contents and extract Aside components in one place. But when importing other markdown files using ESM, the rehype-layout plugin throws esbuild off because `GuideLayout` (etc) doesn't exist except for the primary markdown file that is being compiled.

To fix this, we needed to _invert_ the parser so that the ToC and the Aside components are extracted separately.

### Alternatives considered
https://www.npmjs.com/package/tunnel-rat this seemed like a promising way to collect components such as `<Aside>` and "tunnel" them into a different part of the react layout. However, this doesn't work in SSR b/c it requires a `useEffect` to set the state. The `<Aside>` component was too significant to be client-side rendered only, so an exception was made just for this component: https://github.com/fern-api/fern-platform/blob/ajiang/mdx-sanitize/packages/ui/fern-docs-mdx/src/plugins/rehype-extract-asides.ts.

---

### Sanitizing acorns and brackets

In MDX, `{` and `<` are special. They denote expressions and JSX elements. However, many of our users are importing markdown from other sources, in which `{` and `<` are used extensively for other reasons (i.e. `{denoting, a, list, of, options}` for example: `v0.1.{2—4}`, or describing a math expression: `jon's apples < peter's pears`)

MDX doesn't provide built-in support for sanitizing this. Docusaurus, for example, lists out all the possible errors: https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3 and advises users to add escapes.

Obviously, this is not great for onboarding. Strategies implemented in this PR:

1. [sanitize-mdx-expression](https://github.com/fern-api/fern-platform/blob/ajiang/mdx-sanitize/packages/ui/fern-docs-mdx/src/sanitize/sanitize-mdx-expression.ts#L11) before chucking the markdown into the mdx compiler, run it through the mdx esm plugin up to 100 times. If there's a know issue, try to resolve it using escapes.
2. [remark-sanitize-acorn](https://github.com/fern-api/fern-platform/blob/ajiang/mdx-sanitize/packages/ui/fern-docs-mdx/src/plugins/remark-sanitize-acorn.ts), use a global allowlist to decide which expressions should be escaped. This is aware of ESM statements, so if `export const foo = 'bar'` exists on the page, it will alllowlist `{foo}` for that page.
3. [rehype-acorn-error-boundary](https://github.com/fern-api/fern-platform/blob/ajiang/mdx-sanitize/packages/ui/fern-docs-mdx/src/plugins/rehype-acorn-error-boundary.ts) using an error boundary around all mdx expressions in order to capture any issues that cannot be resolved via static analysis.
4. [extract-jsx](https://github.com/fern-api/fern-platform/blob/ajiang/mdx-sanitize/packages/ui/fern-docs-mdx/src/extract-jsx.ts) capture all jsx elements that are referenced on the page. This list will be used to inject error boundaries at run-time: https://github.com/fern-api/fern-platform/blob/ajiang/mdx-sanitize/packages/ui/app/src/mdx/components/index.tsx#L135-L150

---

### What's next

We'll likely need to fork mdx-bundler to be able to customize how esm dependencies are built:

1. there's a lot of rehype plugins that transform fern-specific components, i.e. CodeBlocks, Accordions, Steps, etc. These may not be compatible when used in ESM imported markdown.
2. push some of the validation upstream (into the CLI)
3. **ensure that the JSX element overrides do not overwrite esm-imported jsx**
4. esm-imported markdown will prevent algolia from indexing them. We need to update the algolia indexer to account for this, and bring in all imported markdown as well. One way we can do this is via a dependency-tracer like: https://github.com/vercel/nft


dsinghvi (approved):
@abvthecity i read through the tests and that made sense to me, im still pretty far from the code here so maybe it makes sense to have @RohinBhargava take a look

PR #1691, opened 2024-10-22
fix: reduce aggressive mdx sanitization

abvthecity:
rehype-sanitize-acorns is over-sanitizing markdown

```
export const cards = ["test"];

cards.map(elem => <div>{elem}</div>);
```

was getting sanitized because `elem` was getting matched.

### DIFF
--- a/packages/ui/app/src/components/FernErrorBoundary.tsx
+++ b/packages/ui/app/src/components/FernErrorBoundary.tsx
@@ -39,23 +39,23 @@ export function FernErrorTag({
 }): ReactElement | null {
     const isLocalPreview = useIsLocalPreview();
     useEffect(() => {
-        // TODO: sentry
-        // eslint-disable-next-line no-console
-        console.error(
-            errorDescription ??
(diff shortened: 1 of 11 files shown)


packages/ui/app/src/components/FernErrorBoundary.tsx
```diff
@@ -39,23 +39,23 @@ export function FernErrorTag({
...
-
-    if (showError || showMarkdownError) {
-        if (fallback != null) {
-            return <>{fallback}</>;
-        }
+    if (fallback != null) {
+        return <>{fallback}</>;
+    }
 
+    // if local preview, always show the error tag for markdown errors
```
RohinBhargava, line 57: was this a bug in logic before? Should this be flipped if not (e.g. `!showError && !isLocalPreview`)

abvthecity: no it wasn't a bug, the idea was that in prod, we don't show error warnings to end users (in this case a tag), but in local preview, we do.

PR #1694, opened 2024-10-22
fix: disconnect after 20 messages to illustrate how websockets work

RohinBhargava:
Fixes FER-3403

- Limits websocket messages to 20 messages maximum in a session, to better serve as an illustration of how a websocket is used.
- Moves websocket messages into a Virtuoso frame, to accommodate infinite scrolling.
- Change log keys to node id, to avoid collisions from other websockets with same endpoint id.

- Customer asks.

https://github.com/user-attachments/assets/a329f91b-19b5-4c67-8801-9ed5bac05543


abvthecity (approved):

packages/ui/docs-bundle/src/server/withInitialProps.ts
```diff
@@ -317,6 +317,7 @@ export async function withInitialProps({
             docs.definition.filesV2,
             node.tabs.length > 0,
         ),
+        playground: docs.definition.config.playground,
```
abvthecity, line 320: this doesn't need to be here because it's already in the navigation config

abvthecity: https://github.com/fern-api/fern-platform/blob/main/fern/apis/fdr/definition/navigation/latest/__package__.yml#L249-L256

PR #1696, opened 2024-10-23
fix: canonical url disambiguation for endpoints between api definitions

abvthecity:
multiple api definitions may have the same endpoint, so this effectively disambiguates the two.


packages/fdr-sdk/src/navigation/migrators/v1ToV2.ts
```diff
@@ -531,10 +531,7 @@ export class FernNavigationV1ToLatest {
     ): FernNavigation.EndpointNode => {
         const slug = FernNavigation.Slug(node.slug);
         const canonicalSlug = this.#getAndSetCanonicalSlug(
-            [
-                `:api:${node.apiDefinitionId}:endpoint:${node.endpointId}`,
-                this.#createTitleDisambiguationKey(node, parents),
-            ],
+            [`:api:endpoint:${node.method}:${node.endpointId}`, this.#createTitleDisambiguationKey(node, parents)],
```
abvthecity, line 534: note: this removes the `{node.apiDefinitionId}` from the disambiguation logic, so the `endpointId` must be consistent across api definitions for this to work.

dsinghvi (approved):

PR #1704, opened 2024-10-24
feat(docs): add edge config for authed previews

dsinghvi:
Add edge config for authed previews. 

Some customers dont want their preview urls to be public.

Unit tests for checking if a domain is a "preview" domain

### DIFF
--- a/packages/ui/docs-bundle/src/server/auth/getAuthState.ts
+++ b/packages/ui/docs-bundle/src/server/auth/getAuthState.ts
@@ -90,10 +90,11 @@ export async function getAuthStateInternal({
                 const workosUserInfo = await toSessionUserInfo(session);
                 if (workosUserInfo.user) {
                     // TODO: should this be stored in the session itself?
+                    const roles = await getWorkosRbacRoles(previewAuthConfig.org, workosUserInfo.user.email);
                     return {
                         authed: true,
                         ok: true,
-                        user: toFernUser(workosUserInfo),
+                        user: toFernUser(workosUserInfo, roles),
                         partner: "workos",
                     };
                 }
--- a/packages/ui/docs-bundle/src/server/auth/getAuthState.ts
+++ b/packages/ui/docs-bundle/src/server/auth/getAuthState.ts
@@ -1,6 +1,6 @@
 import { withDefaultProtocol } from "@fern-api/ui-core-utils";
 import { AuthEdgeConfig, FernUser } from "@fern-ui/fern-docs-auth";
-import { getAuthEdgeConfig } from "@fern-ui/fern-docs-edge-config";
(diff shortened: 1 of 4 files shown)


packages/ui/fern-docs-edge-config/src/getPreviewUrlAuthConfig.ts
```diff
@@ -0,0 +1,33 @@
...
+export async function getPreviewUrlAuthConfig(
+    currentDomain: string,
+): Promise<z.infer<typeof PreviewUrlAuthSchema> | undefined> {
+    const previewDomain = isPreviewDomain(currentDomain);
+    if (!previewDomain) {
+        return undefined;
+    }
+
+    const config = await get<PreviewUrlAuthConfig>("authed-previews");
+    return config?.[currentDomain];
```
dsinghvi, line 27: need to get to org

packages/ui/docs-bundle/src/server/auth/getAuthState.ts
```diff
@@ -72,16 +72,43 @@ export async function getAuthStateInternal({
...
+                const redirectUri = urlJoin(
+                    removeTrailingSlash(withDefaultProtocol(host)),
+                    "/api/fern-docs/auth/sso/callback",
+                );
+                const authorizationUrl = getWorkosSSOAuthorizationUrl({
+                    redirectUri,
+                    organization: previewAuthConfig.org,
+                });
+                return { authed: false, ok: true, authorizationUrl, partner: "workos" };
+            }
```
dsinghvi, line 110: @abvthecity does this look right? wanted to get the logic right before making the code perfect

packages/ui/docs-bundle/src/server/auth/getAuthState.ts
```diff
@@ -72,16 +72,43 @@ export async function getAuthStateInternal({
...
+        if (previewAuthConfig != null) {
+            if (previewAuthConfig.type === "workos") {
+                const session = fernToken != null ? await getSessionFromToken(fernToken) : undefined;
+                const workosUserInfo = await toSessionUserInfo(session);
+                if (workosUserInfo.user) {
+                    // TODO: should this be stored in the session itself?
+                    return {
+                        authed: true,
+                        ok: true,
+                        user: toFernUser(workosUserInfo),
```
abvthecity, line 96: We should still pass in the FGA roles here

abvthecity: (Note that FGA is hard coded to using staging at the moment, so we will need to productionalize in the future)

abvthecity (approved):

packages/ui/docs-bundle/src/server/auth/getAuthState.ts
dsinghvi: cool will add that, i would have that roles dont matter since specifically making sure `authConfig` is null

PR #1706, opened 2024-10-24
feat(docs): implement a get org for url endpoint in nextjs server

dsinghvi:
Adds an endpoint to go from `url -> org id`. This leverages the vercel kv cache. 

We need a fast way to compute this to use the edge config for authed preview urls. 

Open to suggestions here.


packages/ui/docs-bundle/src/pages/api/fern-docs/org/orgForUrl.ts
```diff
@@ -0,0 +1,24 @@
+import { DocsLoader } from "@fern-ui/fern-docs-server";
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse): Promise<void> {
```
abvthecity, line 4: the path of this handler should be `/api/fern-docs/org/org-for-url` not `/api/fern-docs/org/orgForUrl`

packages/ui/docs-bundle/src/pages/api/fern-docs/org/orgForUrl.ts
```diff
@@ -0,0 +1,24 @@
...
+    const { domain } = req.query;
+
+    if (!domain || typeof domain !== "string") {
+        res.status(400).json({ error: "Invalid domain" });
```
abvthecity, line 8: use the `getXFernHostEdge(req)` to grab the domain

packages/ui/fern-docs-server/src/DocsKVCache.ts
```diff
@@ -0,0 +1,44 @@
...
+            // eslint-disable-next-line no-console
+            console.error(`Could not get ${key} from cache`, e);
+            return null;
+        }
+    }
+
+    public async setOrgId(orgId: string): Promise<void> {
+        const key = this.createKey("orgId");
+        try {
+            await kv.set(key, orgId);
```
abvthecity, line 38: should we set an expiration?

PR #1722, opened 2024-10-25
fix: revalidate-all after promotion, and add comment

abvthecity:


clis/vercel-scripts/src/utils/revalidator.ts
```diff
@@ -17,6 +17,7 @@ export class FernDocsRevalidator {
         this.teamId = teamId;
     }
 
+    // TODO: this should not be using vercel as the source of truth for domains becuase vercel's domain list does not include all subpath (proxied) domains.
```
dsinghvi, line 20: @RohinBhargava ideally your endpoint takes care of this as well

dsinghvi (approved):

PR #1729, opened 2024-10-26
chore: migrate websocket proxy worker into fern-platform

abvthecity:


dannysheridan:
completed in another PR

PR #1738, opened 2024-10-28
feat: sso auth integration with workos

abvthecity:
Orchestrates SSO and social connect using workos + authkit


RohinBhargava (approved):
Approved, but can you elaborate on how we will move forward with supporting other SSO providers? Will we introduce new abstractions to disambiguate the code path in subsequent pushes?

abvthecity:
> Approved, but can you elaborate on how we will move forward with supporting other SSO providers? Will we introduce new abstractions to disambiguate the code path in subsequent pushes?

yeah that's correct. eventually this is gonna look more like an abstract class implementation. although i imagine there's some idiosyncrasies between auth providers that will make this annoying.

the config is setup to make it possible to support other auth providers as well. Moving forward, the rule is, at most one auth provider is allowed per docs instance, so if we need to break stuff, we can easily quarantine a few domains.

PR #1746, opened 2024-10-29
fix: redirect auth callback for app-staging

abvthecity:


packages/ui/docs-bundle/src/pages/api/fern-docs/auth/sso/callback.ts
```diff
@@ -40,11 +34,25 @@ export default async function handler(req: NextRequest): Promise<NextResponse> {
...
+            FORWARDED_HOST_QUERY,
+            "is the same as the host:",
+            String(req.nextUrl.searchParams.get(FORWARDED_HOST_QUERY)),
+        );
+        return new NextResponse(null, { status: 400 });
+    }
+
     // TODO: this is a security risk (open redirect)! We need to verify that the target host is one of ours.
-    if (getHostEdge(req) !== url.host) {
+    // if the current url is app.buildwithfern.com, we should redirect to ***.docs.buildwithfern.com
```
dsinghvi, line 51: do we need to handle dev ? `app-dev.buildwithfern.com` -> `***.docs.dev.buildwithfern.com`

dsinghvi (approved):

packages/ui/docs-bundle/src/pages/api/fern-docs/auth/sso/callback.ts
abvthecity: the comment is illustrating an example— the real domain is stored in `url.host`

PR #1748, opened 2024-10-30
fix: logout url for sso

abvthecity:
Before: the logoutUrl generated by workos doesn't delete cookies, and it always redirects to buildwithfern.com

### DIFF
--- a/packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
+++ b/packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
@@ -1,4 +1,4 @@
-import { getLogoutUrl } from "@/server/auth/workos-session";
+import { revokeSessionForToken } from "@/server/auth/workos-session";
 import { safeUrl } from "@/server/safeUrl";
 import { getDocsDomainEdge, getHostEdge } from "@/server/xfernhost/edge";
 import { withDefaultProtocol } from "@fern-api/ui-core-utils";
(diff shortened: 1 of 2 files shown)


packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
```diff
@@ -13,14 +13,12 @@ export default async function GET(req: NextRequest): Promise<NextResponse> {
...
-    const logoutUrlRaw =
-        authConfig?.type === "basic_token_verification"
-            ? authConfig.logout
-            : authConfig?.partner === "workos"
-              ? await getLogoutUrl(req.cookies.get(COOKIE_FERN_TOKEN)?.value)
-              : undefined;
+    if (authConfig?.type === "sso" && authConfig.partner === "workos") {
+        // revoke session in WorkOS
+        await revokeSessionForToken(req.cookies.get(COOKIE_FERN_TOKEN)?.value);
+    }
```
dsinghvi, line 19: should we do an assert never on all types of authConfigs ? ie if i were to add username/password?

packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
```diff
@@ -13,14 +13,12 @@ export default async function GET(req: NextRequest): Promise<NextResponse> {
...
 
-    const logoutUrl = safeUrl(logoutUrlRaw);
+    const logoutUrl = safeUrl(authConfig?.type === "basic_token_verification" ? authConfig.logout : undefined);
```
dsinghvi, line 21: ideally would make sense to put in a helper function and have it switch through all cases and return a logout url imo

dsinghvi (commented):
small comments, defer to you if it makes sense to implement

dsinghvi (approved):

packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
abvthecity: currently this function is self-contained and the logout behavior (as you can see with this PR) is handled differently based on the auth provider.

packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
abvthecity: lets do that when you do add username/password?

PR #1749, opened 2024-10-31
fix(fdr): filter to latest SDK row and add logging

dsinghvi:
Make sure that when we filter to latest SDK, we don't include "undefined".

Snippets are going missing. 

WIll be tested on dev + existing tests pass.


servers/fdr/src/db/sdk/SdkDao.ts
```diff
@@ -171,13 +172,13 @@ export class SdkDaoImpl implements SdkDao {
             },
             where: {
                 package: sdkPackage,
-                version,
```
dsinghvi, line 174: this can be undefined which would cause this to fail

PR #1751, opened 2024-10-31
fix: correctly delete logout cookie

abvthecity:
1) fern_token cookie isn't being deleted upon logout because we're not setting its `domain` and `path` values.
2) this pr fixes how the `getHostNode` and `getHostEdge` works:

- when the user is loading docs from `docs.domain.com`, the host should be `docs.domain.com`
- when the user is loading docs from `buildwithfern.com/learn` the host should be `buildwithfern.com` (even though it's actually proxied to app.buildwithfern.com)
- when the user is loading docs from a preview or staging url, such as `x.docs.staging.buildwithfern.com`, the host should be `x.docs.staging.buildwithfern.com`
- when the user is loading docs via a PR preview link, the host is still `xxx.vercel.app` because the `__fern_preview_host` cookie should be ignore.

This is done via
1. check if x-fern-host header exists. If so, use it (because this indicates that the url is proxied)
2. default to `host` header, or the URL host, if it exists (which is typically the origin interacting with vercel)

### DIFF
--- a/packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
+++ b/packages/ui/docs-bundle/src/pages/api/fern-docs/auth/logout.ts
@@ -1,3 +1,4 @@
+import { withDeleteCookie } from "@/server/auth/with-secure-cookie";
 import { revokeSessionForToken } from "@/server/auth/workos-session";
 import { safeUrl } from "@/server/safeUrl";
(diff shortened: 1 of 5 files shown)


RohinBhargava (commented):
Technically, when logging out, we might want to issue a token revocation to the issuing server. I'm not sure if this logic is already captured somewhere else, but looks good to me otherwise

RohinBhargava (approved):

PR #1753, opened 2024-10-31
fix: Properly render errors on endpoints

RohinBhargava:
Fixes FER-3459

* This is a partial rewrite of the API reference error rendering logic.
* Now we use a top level object to pass example components to (ideally in the future) pure(r) components downstream, based on an accessor pattern given selected variables. This shape is:
```
{
  [language: string (clientLanguage)]: {
    [requestExample: string (unique key for request)]: {
      [statusCode: number]: {
        examples: CodeExample[]
      }
    }
  }
}
```
* This couples with `selectedError` for bidirectional control flow from shared components. Ideally these are converged in a re-write.
* `endpoint.errors` are merged with `endpoint.examples` to render appropriately. This is quite messy, consider doing this more cleanly upstream.
* As discussed, we believe that the key for showing examples should be based on `Request`, and show the appropriate responses, bifurcating on the examples in a given list. This is too much to accomplish in this time frame.

* Broken renderings in many places in various endpoint docs.

https://github.com/user-attachments/assets/b958f98c-6e4b-494a-a3d8-f02112ef726b

https://github.com/user-attachments/assets/1deafefe-d08c-4366-a4bb-52e105b7c8ec

https://github.com/user-attachments/assets/56a89c78-52dd-4e31-985b-6ee4f12ddd66


abvthecity:
Screenshot 2024-10-31 at 5 13 54 PM

@RohinBhargava there's an extra line at the top can you remove?

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
```diff
@@ -155,6 +150,179 @@ export const EndpointContent = memo<EndpointContent.Props>((props) => {
...
+    useEffect(() => {
+        setSelectedExampleKey([
+            selectedClient.language,
+            selectedClient.key,
+            selectedClient.exampleCall.responseStatusCode,
+            0,
+        ]);
+    }, [selectedClient]);
+
+    const convertErrorResponseToCodeExamples = useMemo(
```
abvthecity, line 170: useCallback

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
```diff
@@ -155,6 +150,179 @@ export const EndpointContent = memo<EndpointContent.Props>((props) => {
...
+                    return [selectedLanguage, exampleId, statusCode, exampleIndex];
+                }
+            }
+
+            return selectedExampleKey;
+        });
+    }, [selectedLanguage, examplesByClientAndTitleAndStatusCode]);
+
+    const handleSelectError = useMemo(
+        () => (error: ApiDefinition.ErrorResponse | undefined) => {
```
abvthecity, line 306: useCallback

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
```diff
@@ -155,6 +150,179 @@ export const EndpointContent = memo<EndpointContent.Props>((props) => {
...
+                                  ),
+                              },
+                          ];
+                      }),
+                  )
+                : {};
+
+            const allGlobalErrors = endpoint.errors
+                ? Object.fromEntries(
+                      Object.entries(groupBy(endpoint.errors, (e) => e.statusCode)).map(
```
abvthecity, line 229: https://es-toolkit.slash.page/reference/object/mapValues.html#mapvalues will probably make this easier to read

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
```diff
@@ -155,6 +150,179 @@ export const EndpointContent = memo<EndpointContent.Props>((props) => {
...
+                    : [];
+            },
+        [endpoint.path],
+    );
+
+    const examplesByClientAndTitleAndStatusCode = useMemo(() => {
+        return clients.reduce<ExamplesByClientAndTitleAndStatusCode>((acc, client) => {
+            acc[client.language] = client.examples
+                ? Object.fromEntries(
+                      Object.entries(
```
abvthecity, line 204: mapValues here too

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
```diff
@@ -155,6 +150,179 @@ export const EndpointContent = memo<EndpointContent.Props>((props) => {
...
+                          groupBy(
+                              client.examples.filter(
+                                  (e) =>
+                                      e.exampleCall.responseStatusCode >= 200 && e.exampleCall.responseStatusCode < 300,
+                              ),
+                              (e) => e.key ?? "",
```
abvthecity, line 210: in what cases will key be undefined?

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
abvthecity: one way to actually avoid having to define a bunch of useMemo/useCallback is to just move this into a separate util file. It doesn't look like this callback needs to be defined within the component itself.

packages/ui/app/src/api-reference/endpoints/EndpointContent.tsx
```diff
@@ -155,6 +150,179 @@ export const EndpointContent = memo<EndpointContent.Props>((props) => {
...
 
+    // We use a string here with the intention that this can be used in a query param to deeplink to a particular example
+    const [selectedExampleKey, setSelectedExampleKey] = useState<SelectedExampleKey | undefined>([
+        selectedClient.language,
+        selectedClient.key,
+        selectedClient.exampleCall.responseStatusCode,
+        0,
+    ]);
+
+    useEffect(() => {
```
abvthecity, line 161: this feels like a somewhat unnecessary useEffect. usually you don't want to introduce an effect like this because it introduces two re-render cycles and can lead to bugs.

abvthecity (approved):
approved in theory, some comments!

PR #1755, opened 2024-10-31
fix: return sdk id that always has snippets

dsinghvi:
PIck an SDK ID that has snippets.  

There are chances of having SDK IDs that do not have associated snippets, causing snippets to disappear. 

Will test on dev.

### DIFF
--- a/servers/fdr/src/db/sdk/SdkDao.ts
+++ b/servers/fdr/src/db/sdk/SdkDao.ts
@@ -166,7 +166,8 @@ export class SdkDaoImpl implements SdkDao {
             }
         }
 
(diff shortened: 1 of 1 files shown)


servers/fdr/src/db/sdk/SdkDao.ts
```diff
@@ -177,7 +178,25 @@ export class SdkDaoImpl implements SdkDao {
             orderBy: {
                 createdAt: "desc",
             },
+            take: 10,
```
dsinghvi, line 181: we are now executing up to 10 more db calls as part of `fern register --docs` for every corrupted SDK. not sure if a join is faster

amckinney (approved):

PR #1761, opened 2024-11-01
found needle in haystack that is overwriting name

RohinBhargava:
Fixes FER-3461 FER-3462

* Does not set name for example snippet for js or ts snippets in FDR. Example generation prefers snippet names, so this aligns with other languages

* Customer reported bugs

* To be tested on staging site (on Mercoa)


packages/fdr-sdk/src/api-definition/migrators/v1ToV2.ts
```diff
@@ -673,7 +673,7 @@ export class ApiDefinitionV1ToLatest {
             codeExamples.typescriptSdk != null
         ) {
             push(jsLang, {
-                name: `${jsLangName} SDK`,
+                name: undefined,
```
RohinBhargava, line 676: Change here

dsinghvi (approved):

PR #1803, opened 2024-11-08
fix: authed previews in production redirect to the CDN URI

dsinghvi:
In production, we should redirect to `app.buildwithfern.com` 

Make sure redirect URIs work in production WorkOS auth

N/A

### DIFF
--- a/packages/ui/docs-bundle/src/server/auth/workos-handler.ts
+++ b/packages/ui/docs-bundle/src/server/auth/workos-handler.ts
@@ -1,0 +1,72 @@
+import { withDefaultProtocol } from "@fern-api/ui-core-utils";
+import { removeTrailingSlash } from "next/dist/shared/lib/router/utils/remove-trailing-slash";
+import urlJoin from "url-join";
+import { AuthState, getWorkosRbacRoles } from "./getAuthState";
+import { getWorkosSSOAuthorizationUrl } from "./workos";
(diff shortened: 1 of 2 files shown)


packages/ui/docs-bundle/src/server/auth/workos-handler.ts
```diff
@@ -0,0 +1,72 @@
...
+                ok: true,
+                user: toFernUser(await toSessionUserInfo(updatedSession), roles),
+                partner: "workos",
+            };
+        }
+    }
+
+    const redirectUri = process.env.NEXT_PUBLIC_CDN_URI != null 
+        ? urlJoin(removeTrailingSlash(withDefaultProtocol(process.env.NEXT_PUBLIC_CDN_URI)), "/api/fern-docs/auth/sso/callback")
+        : urlJoin(removeTrailingSlash(withDefaultProtocol(host)), "/api/fern-docs/auth/sso/callback");
```
dsinghvi, line 63: this is the key logic change -- i added it for all callers, is that right? @abvthecity

abvthecity (approved):

packages/ui/docs-bundle/src/server/auth/workos-handler.ts
abvthecity: I would do something like this:
```
String(new URL("/api/fern-docs/auth/sso/callback", withDefaultProtocol(process.env.NEXT_PUBLIC_CDN_URI ?? host)))
```

PR #1814, opened 2024-11-14
feat: foundation for OpenApi Parser

RohinBhargava:
This PR introduces the base nodes that are necessary for parsing OpenApi Spec.

It introduces 2 abstractions:
* BaseAPIConverterNode: This defines the domain model for the parser. It contains a Context, containing an ErrorCollector and a Logger. It defines an interface for how to model outputting FDR shapes defined as generics for classes.
* BaseAPIOpenApiV3_1Node: These extend from BaseAPIConverterNode map to OpenApi specific nodes, and can be incrementally updated, so that pathing and location is maintained.

The nodes created are:
* String --> these correspond to any type: "string" nodes we encounter in OpenApi spec
* Number --> these correspond to any type: "integer | number" nodes we encounter in OpenApi spec
  - Integer: These map to "integer" type nodes and handle discrimination into the right FDR shapes.
  - Number: These map to "float" type nodes and handle discrimination into the right FDR shapes.
* Schema: An input node that helps us recognize schemas/references as we encounter them in OpenApi spec.
* Reference: An input node that helps us to resolve references or pass them as Alias types.
* Object (and all derivatives): An output node that helps us to parse any schema from OpenApi into the correct Fdr shapes.

Additionally, there are a few of types files introduced for cleaner organization:
* OpenApi formats: this keeps an exhaustive inventory of potential formats encountered on different primitives
* FdrShapes: Groups FdrShapes that are closely related, so that top level nodes can share logic for all children


packages/parsers/openapi/shared/composable/availability.node.ts
```diff
@@ -0,0 +1,35 @@
...
+        ComposableApiNode<
+            T,
+            InputNode,
+            T & {
+                availability: string;
+            }
+        >
+{
+    name = "availability";
+    preProcessedInput: InputNode["preProcessedInput"];
```
dsinghvi, line 14: what is `preProcessedInput` ?

packages/parsers/openapi/shared/interfaces/api.node.interface.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export interface ApiNode<InputShape, FdrShape> {
+    context: ApiNodeContext;
+    preProcessedInput: InputShape;
+
+    accessPath: ApiNode<unknown, unknown>[];
+    name: string;
+
+    qualifiedId: string;
```
dsinghvi, line 16: what is a `qualifiedId` ?

packages/parsers/openapi/shared/stages/fdr/api.stage.ts
```diff
@@ -0,0 +1,40 @@
...
+    endpoints: ApiProcessingStageOutput["endpoints"] = {};
+    websockets: ApiProcessingStageOutput["websockets"] = {};
+    webhooks: ApiProcessingStageOutput["webhooks"] = {};
+
+    constructor(
+        readonly context: ApiNodeContext,
+        readonly preProcessedInput: OpenAPIV3_1.Document["paths"],
+        readonly accessPath: ApiNode<unknown, unknown>[],
+    ) {
+        this.qualifiedId = `${this.accessPath.map((node) => node.qualifiedId).join(".")}.${this.name}`;
```
dsinghvi, line 26: seeing this line repeated quite a bit -- thoughts on refactoring this logic to a base abstract class ?

packages/parsers/openapi/shared/interfaces/api.node.interface.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    outputFdrShape: () => FdrShape;
```
dsinghvi, line 18: Not sure if `WithoutQuestionMarks<FdrShape>` will work, but it would be great to force every `ApiNode` to set every key/value pair

packages/parsers/openapi/shared/composable/availability.node.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    outputFdrShape(): T & {
+        availability: string;
+    } {
+        const baseShape = this.inputNode.outputFdrShape();
+        return {
+            ...baseShape,
+            availability: this.inputNode.availability,
+        };
+    }
```
dsinghvi, line 34: Thoughts on how you are planning to handle errors/excdeptions as you do the conversion? 

It could be helpful if every output FdrShape was wrapped in a discriminated union that had a success/error return value. This might be nice so that even if someone has a misconfigured part of their OpenAPI spec, you dont have to fail hard. 

```
ApiOutput = SuccessApiOutput | ErrorApiOutput; 

interface SuccessApiOutput {
  type: "success"; 
  result: FdrShape
}

interface ErrorOutput {
  type: "error";
  message: string;
}
```

dsinghvi: Another option could be creating an `ErrorCollector` where you can collect validation errors as part of the converter.

packages/parsers/openapi/shared/interfaces/fdr.stage.interface.ts
```diff
@@ -0,0 +1,5 @@
+import { ApiNode } from "./api.node.interface";
+
+export interface FdrStage<InputShape, FdrShape> extends ApiNode<InputShape, FdrShape> {
```
dsinghvi, line 3: i think it might be helpful to have some basic docs about what each primitive is (Stage vs. Node vs. ContainerNode).

packages/parsers/openapi/shared/composable/availability.node.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    outputFdrShape(): T & {
+        availability: string;
+    } {
+        const baseShape = this.inputNode.outputFdrShape();
+        return {
+            ...baseShape,
+            availability: this.inputNode.availability,
+        };
```
dsinghvi, line 33: im a little confused how this works -- we parse availability from `x-fern-availability` or using the `deprecated: true/false` in OpenAPI. How does `this.inputNode.availability` start coming through

packages/parsers/openapi/shared/composable/availability.node.ts
RohinBhargava: So these are intended to be the OpenApi Document shaped objects -- probably can have a better name here, like openApiShape?

packages/parsers/openapi/shared/interfaces/api.node.interface.ts
```diff
@@ -0,0 +1,26 @@
+import { Logger } from "@playwright/test";
+
+export interface ApiNodeContext {
+    orgId: string;
+    apiId: string;
+    logger: Logger;
```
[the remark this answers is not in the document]
RohinBhargava: yes, definitely, will bring in the logger dependency. This is just a placeholder for illustrative purposes.

packages/parsers/openapi/shared/interfaces/api.node.interface.ts
RohinBhargava: qualifiedId is intended to be a unique id for a node in the path descent, so that it's easy to debug if we need to use the node anywhere else in a processing step

packages/parsers/openapi/shared/stages/fdr/api.stage.ts
RohinBhargava: yeah, agreed -- was thinking there might be some differences in the way these are handled, but after doing a few nodes, definitely makes sense to push back into the abstraction and have overrides if we need to

packages/parsers/openapi/shared/interfaces/api.node.interface.ts
RohinBhargava: ah yeah, we can do that for sure

packages/parsers/openapi/shared/composable/availability.node.ts
RohinBhargava: I think the error collector makes sense -- agreed on not failing hard -- we will have to stop the processing for any of the applicable descendants, but we should definitely get as wide of a parsing view as possible

packages/parsers/openapi/shared/interfaces/fdr.stage.interface.ts
RohinBhargava: makes sense, I will add them as jsdoc comments so that they can be read in code too

packages/parsers/openapi/shared/composable/availability.node.ts
RohinBhargava: `this.inputNode.availability` should actually be `this.inputNode["x-fern-availability"]` in this case. My basic assumption is that the object being passed will carry extensions, even if the OpenApi Document shape does not allow for it. Above in the top level signature, we can define the extension key and shape that we expect (in the next iteration of this PR, will demonstrate with 'x-fern-availability')

packages/parsers/openapi/shared/composable/availability.node.ts
dsinghvi: Even if you have an error collector the return value needs to be `undefined` or OutputShape

packages/parsers/openapi/shared/composable/availability.node.ts
dsinghvi: `this.inputNode["x-fern-availability"]` that makes sense, I would change the file hierarchy then and put this in an `extensions` directory. 

Also how are you going to handle `deprecated: true` or `deprecated: false` ?

packages/parsers/openapi/base.node.interface.ts
```diff
@@ -0,0 +1,49 @@
+import { Logger } from "@playwright/test";
+
+export class ErrorCollector {
+    errors: {
+        message: string;
+        path: string;
```
dsinghvi, line 6: up to export the `ValidationError` as a separate type? and can we model as `path: string[]` so that consumers can render it however they want

packages/parsers/openapi/base.node.interface.ts
```diff
@@ -0,0 +1,49 @@
...
+    }[] = [];
+
+    addError(error: string, accessPath: string[], pathId?: string): void {
```
dsinghvi, line 9: would love to keep this as simple as possible, can we delete `pathId?: string` ?

packages/parsers/openapi/base.node.interface.ts
```diff
@@ -0,0 +1,49 @@
...
+        this.errors.push({
+            message: error,
+            path: `#/${accessPath.join("/")}${pathId ? `/${pathId}` : ""}`,
+        });
+    }
+}
+
+export interface ApiNodeContext {
```
dsinghvi, line 17: nit: can we move `ApiNodeContext`, and `ApiNode` into a separate file called `ApiNode.ts` ?

packages/parsers/openapi/base.node.interface.ts
```diff
@@ -0,0 +1,49 @@
+import { Logger } from "@playwright/test";
+
+export class ErrorCollector {
```
dsinghvi, line 3: nit: can we move `ErrorCollector` into a `ErrorCollector.ts`

packages/parsers/openapi/shared/nodes/object.node.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export class ObjectNode extends OutputApiNode<SchemaObject, FdrAPI.api.latest.ObjectType> {
+    extends: FdrAPI.TypeId[] = [];
+    properties: ObjectPropertyNode[] = [];
+    extraProperties: TypeReferenceNode | undefined;
+
+    constructor(context: ApiNodeContext, input: SchemaObject, accessPath: string[], accessorKey?: string) {
+        super(context, input, accessPath);
+
+        if (input.allOf !== undefined) {
```
dsinghvi, line 15: nit: generally prefer `!= null` checks to handle both undefined and null

packages/parsers/openapi/shared/nodes/object.node.ts
```diff
@@ -0,0 +1,40 @@
...
+            this.extends = input.allOf
+                .map((type) => (isReferenceObject(type) ? FdrAPI.TypeId(type.$ref) : undefined))
+                .filter((id): id is FdrAPI.TypeId => id !== undefined);
```
dsinghvi, line 18: there should be a helper called `isNonNullish` that should allow you to skip writing the type guard

packages/parsers/openapi/base.node.interface.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    accessPath: string[];
+
+    constructor(context: ApiNodeContext, input: InputShape, accessPath: string[]) {
+        this.context = context;
+        this.input = input;
+        this.accessPath = accessPath;
+    }
+
+    abstract outputFdrShape: () => FdrShape | undefined;
```
dsinghvi, line 36: nit: rename to `toFdrShape` ?

packages/parsers/__test__/shared/nodes/object.node.test.ts
```diff
@@ -0,0 +1,87 @@
+import { FdrAPI } from "@fern-api/fdr-sdk";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { ObjectNode } from "../../../openapi/shared/nodes/object.node";
+import { SchemaObject } from "../../../openapi/shared/openapi.types";
+import { createMockContext } from "../../createMockContext.util";
+
+describe("ObjectNode", () => {
```
dsinghvi, line 7: mega nit: i think it might be nicer to put these test files through out the code base as opposed to the top, that way you can see `object.node.ts` and `object.node.test.ts` side-by-side

packages/parsers/openapi/shared/nodes/object.node.ts
```diff
@@ -0,0 +1,39 @@
...
+    extends: FdrAPI.TypeId[] = [];
+    properties: ObjectPropertyNode[] = [];
+    extraProperties: TypeReferenceNode | undefined;
+
+    constructor(context: ApiNodeContext, input: SchemaObject, accessPath: string[], accessorKey?: string) {
+        super(context, input, accessPath);
+
+        if (input.allOf != null) {
+            this.extends = input.allOf
+                .map((type) => (isReferenceObject(type) ? FdrAPI.TypeId(type.$ref) : undefined))
```
dsinghvi, line 18: if a `$ref` is `#/components/schemas/Person`, ideally we store a TypeId that is `Person`

packages/parsers/openapi/shared/nodes/primitives/enum.node.ts
```diff
@@ -0,0 +1,38 @@
+import { FdrAPI } from "@fern-api/fdr-sdk";
+import { OpenAPIV3_1 } from "openapi-types";
+import { ApiNodeContext, InputApiNode } from "../../../ApiNode";
+
+export class EnumNode extends InputApiNode<OpenAPIV3_1.SchemaObject, FdrAPI.api.v1.read.PrimitiveType.Enum> {
```
dsinghvi, line 5: nit: think we should move enum outside of primitives folder

packages/parsers/openapi/shared/nodes/primitives/number/float.node.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        if (input.type !== "number") {
+            context.errorCollector.addError(
+                `Expected type "number" for numerical primitive, but got "${input.type}"`,
+                accessPath,
+            );
+            return;
+        }
+
+        if (!isOpenApiNumberTypeFormat(input.format)) {
```
dsinghvi, line 25: i don't think this actually needs to be the case, i.e. you can have numbers without formats

packages/parsers/openapi/shared/nodes/primitives/number/float.node.ts
```diff
@@ -0,0 +1,57 @@
...
+
+        switch (input.format) {
+            case "decimal":
+            case "decimal128":
+            case "double-int":
+            case "double":
+            case "float":
+            case "sf-decimal":
+            case undefined:
+                this.type = "double";
```
dsinghvi, line 41: note: we'll eventually want to update FDR to store the actual format

packages/parsers/openapi/shared/nodes/primitives/number/integer.node.ts
```diff
@@ -0,0 +1,56 @@
...
+            );
+            return;
+        }
+
+        if (!isOpenApiIntegerTypeFormat(input.format)) {
+            context.errorCollector.addError(
+                `Expected format for integer primitive, but got "${input.format}"`,
+                accessPath,
+            );
+            return;
```
dsinghvi, line 30: same comment as flat, dont think this need to be a large failure

packages/parsers/openapi/ErrorCollector.ts
```diff
@@ -0,0 +1,14 @@
+type ValidationError = {
+    message: string;
+    path: string[];
+};
+export class ErrorCollector {
+    errors: ValidationError[] = [];
+
+    addError(error: string, accessPath: string[]): void {
```
dsinghvi, line 8: I think it also makes sense to collect `warnings` + `errors`, 

Could see an example of `collector.warn()` and `collector.error()` as the two methods on the ErrorCollector.

packages/parsers/openapi/shared/openapi.types.ts
```diff
@@ -0,0 +1,3 @@
+import { OpenAPIV2, OpenAPIV3, OpenAPIV3_1 } from "openapi-types";
+export type SchemaObject = OpenAPIV3_1.SchemaObject | OpenAPIV3.SchemaObject | OpenAPIV2.SchemaObject;
+export type ReferenceObject = OpenAPIV3_1.ReferenceObject | OpenAPIV3.ReferenceObject | OpenAPIV2.ReferenceObject;
```
dsinghvi, line 3: can we delete this (since we are only optimizing for OpenAPIV3_1 for now) ?

packages/parsers/openapi/3.1/schemas/ObjectConverter.node.ts
```diff
@@ -0,0 +1,80 @@
...
+import { ObjectPropertyConverterNode } from "./ObjectPropertyConverter.node";
+import { SchemaConverterNode } from "./SchemaConverter.node";
+
+export declare namespace ObjectConverterNode {
+    interface Input extends OpenAPIV3_1.NonArraySchemaObject {
+        type: "object";
+    }
+}
+
+function convertExtraPropertiesBoolean(displayName: string | undefined): {
```
RohinBhargava, line 15: this pattern is interesting -- it is done for extensibility. If we want to add a new type of output to a node in the future, can be as simple as adding a new member to the return type of this function. @dsinghvi

packages/parsers/openapi/3.1/schemas/BooleanConverter.node.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    constructor(
+        input: BooleanConverterNode.Input,
+        context: BaseAPIConverterNodeContext,
+        accessPath: string[],
+        pathId: string,
+    ) {
+        super(input, context, accessPath, pathId);
+
+        if (typeof input.default !== "boolean") {
```
dsinghvi, line 26: what if it equals null, wont this error?

packages/parsers/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -0,0 +1,130 @@
...
+import { StringConverterNode } from "./StringConverter.node";
+
+type PrimitiveType =
+    | NumberConverterNode.Input
+    | IntegerConverterNode.Input
+    | BooleanConverterNode.Input
+    | StringConverterNode.Input;
+
+export class SchemaConverterNode extends BaseOpenApiV3_1Node<
+    OpenAPIV3_1.SchemaObject,
```
dsinghvi, line 18: i think this probably should handle Schema | Reference

packages/parsers/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -0,0 +1,130 @@
...
+        //     break;
+        // } else if (input.oneOf != null) {
+        //     this.typeNode = new OneOfConverterNode(input, this.context, this.accessPath);
+        //     break;
+        // } else if (input.allOf != null) {
+        //     this.typeNode = new AllOfConverterNode(input, this.context, this.accessPath);
+        //     break;
+        // }
+
+        if (input != null) {
```
dsinghvi, line 62: can the input here even be null?

packages/parsers/openapi/3.1/schemas/StringConverter.node.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export class StringConverterNode extends BaseOpenApiV3_1Node<StringConverterNode.Input, FdrStringType> {
+    type: FdrStringType["type"] | undefined;
+    regex: string | undefined;
+    default: string | undefined;
+    minLength: number | undefined;
+    maxLength: number | undefined;
+
+    private mapToFdrType = (
+        format: ConstArrayToType<typeof OPENAPI_STRING_TYPE_FORMAT>,
```
dsinghvi, line 26: what happens if format is undefined

packages/parsers/openapi/utils/chooseReferenceObjectOrSchema.ts
```diff
@@ -0,0 +1,17 @@
...
+    input: OpenAPIV3_1.ReferenceObject | OpenAPIV3_1.SchemaObject,
+    context: BaseAPIConverterNodeContext,
+    accessPath: string[],
+    pathId: string,
+): SchemaConverterNode | ReferenceConverterNode | undefined {
+    if (isReferenceObject(input)) {
+        return new ReferenceConverterNode(input, context, accessPath, pathId);
+    }
+    return new SchemaConverterNode(input, context, accessPath, pathId);
+}
```
dsinghvi, line 17: nit: thoughts on putting this in the `SchemaNode` that way regardless of whatever schema you have: a reference or an inlined shape, it will work

packages/parsers/openapi/utils/mapReferenceObject.ts
```diff
@@ -0,0 +1,5 @@
+import { OpenAPIV3_1 } from "openapi-types";
+
+export function mapReferenceObject(referenceObject: OpenAPIV3_1.ReferenceObject): OpenAPIV3_1.SchemaObject {
+    return referenceObject.$ref as OpenAPIV3_1.SchemaObject;
```
dsinghvi, line 4: this feels suspect, what's the utility of this coercion?

packages/parsers/openapi/3.1/schemas/ObjectConverter.node.ts
```diff
@@ -0,0 +1,80 @@
...
+        super(input, context, accessPath, pathId);
+
+        this.extends = [];
+
+        // TODO: discriminate between schemas and references. For schemas, we need to inline the properties for processing
+
+        // if (input.allOf != null) {
+        //     this.extends = input.allOf
+        //         .map((type) => chooseReferenceOrSchemaNode(type, context, accessPath, pathId))
+        //         .filter(isNonNullish);
```
dsinghvi, line 52: what does `chooseReferenceOrSchemaNode` do? I could see a `SchemaOrReferenceConverterNode.ts` so that you can just directly pass the input in

packages/parsers/openapi/3.1/schemas/ObjectPropertyConverter.node.ts
```diff
@@ -0,0 +1,56 @@
...
+
+        // unique to OAS 3.1, properties can also live alongside $ref, so we will use this to determine the description and availability
+        this.description = input.description;
+
+        // TODO: add availability
+        // this.availability = input.availability;
+
+        if (this.valueShape == null) {
+            this.context.errors.error({
+                message: `Unprocessable type shape: ${this.propertyKey}`,
```
dsinghvi, line 37: errors should speak the language of the input to make it easier for the user to understand

packages/parsers/openapi/3.1/schemas/ObjectPropertyConverter.node.ts
```diff
@@ -0,0 +1,56 @@
+import { FdrAPI } from "@fern-api/fdr-sdk";
+import { OpenAPIV3_1 } from "openapi-types";
+import { BaseAPIConverterNodeContext } from "../../../BaseApiConverter.node";
+import { BaseOpenApiV3_1Node } from "../../BaseOpenApiV3_1Converter.node";
+import { AvailabilityNode } from "../../shared/temporary/demo";
+import { chooseReferenceOrSchemaNode } from "../../utils/chooseReferenceObjectOrSchema";
+import { ReferenceConverterNode } from "./ReferenceConverter.node";
+import { SchemaConverterNode } from "./SchemaConverter.node";
+
+export class ObjectPropertyConverterNode extends BaseOpenApiV3_1Node<
```
dsinghvi, line 10: why do you need this entire class at all? ideally you should just be able to redirect all object properties back to the schema node (example: https://github.com/fern-api/fern/blob/5ab95904f1a1d1a0fb4a950d1f046b8d603b78c6/generators/openapi/src/converters/convertObject.ts#L30)

packages/parsers/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -0,0 +1,130 @@
...
+    | NumberConverterNode.Input
+    | IntegerConverterNode.Input
+    | BooleanConverterNode.Input
+    | StringConverterNode.Input;
+
+export class SchemaConverterNode extends BaseOpenApiV3_1Node<
+    OpenAPIV3_1.SchemaObject,
+    FdrAPI.api.latest.TypeShape | undefined
+> {
+    typeShapeNode: { convert: () => FdrAPI.api.latest.TypeShape | undefined } | undefined;
```
dsinghvi, line 21: what is this?

packages/parsers/openapi/3.1/schemas/ObjectConverter.node.ts
dsinghvi: cool -- it feels a little complex but happy to defer to you

packages/parsers/openapi/3.1/schemas/SchemaConverter.node.ts
RohinBhargava: This is the same pattern as above, it encapsulates the conversion function so that it's easy to reason about how the domain object functions should convert -- if singular functions, those will be splayed into the constructor (just a way to group for different exports)

packages/parsers/openapi/3.1/schemas/ObjectPropertyConverter.node.ts
RohinBhargava: this is a good point, this logic should live in schema node, yeah

packages/parsers/openapi/3.1/schemas/ObjectConverter.node.ts
RohinBhargava: discussed offline -- this might be confusing for folks if there are two types of nodes that accept ReferenceObjects, but is also easily portable into the SchemaNode

packages/parsers/openapi/utils/mapReferenceObject.ts
RohinBhargava: oh, this is a placeholder for now -- need to migrate the resolution logic in from fern main repo, and it should live in this function. I'll leave a TODO comment

packages/parsers/openapi/3.1/schemas/SchemaConverter.node.ts
RohinBhargava: no, this was just to satisfy the type checker in an intermediate fashion, but can be removed

packages/parsers/openapi/3.1/schemas/StringConverter.node.ts
RohinBhargava: it returns "string". The property should be set to default to "string", however

packages/parsers/types/format.types.ts
```diff
@@ -0,0 +1,50 @@
...
+    "sf-binary",
+    "sf-boolean",
+    "sf-string",
+    "sf-token",
+    "time",
+    "uri-reference",
+    "uri-template",
+    "uri",
+    "uuid",
+] as const;
```
dsinghvi, line 50: nit: should this live under `openapi` folder?

packages/parsers/openapi/3.1/OpenApiV3_1Converter.node.ts
```diff
@@ -0,0 +1,58 @@
...
+//             endpoints,
+//             websockets,
+//             webhooks,
+//             types,
+//             subpackages: {},
+//             auths,
+//             globalHeaders,
+//         };
+//     };
+// }
```
dsinghvi, line 58: nit: can we delete this class ahead of merging this?

packages/parsers/__test__/createMockContext.util.ts
```diff
@@ -0,0 +1,12 @@
+import { createLogger } from "@fern-api/logger";
+import { BaseAPIConverterNodeContext } from "../BaseApiConverter.node";
+import { ErrorCollector } from "../ErrorCollector";
+
+export function createMockContext(): BaseAPIConverterNodeContext {
+    return {
+        orgId: "orgId",
+        apiId: "apiId",
```
dsinghvi, line 8: Do we need `orgId` + `apiId`? Would be great to remove this from the context so that folks can leverage without having to specify

dsinghvi: I've learned the hard way that these required fields become really hard to remove after the fact

PR #1828, opened 2024-11-18
feat: llms.txt

abvthecity:
Introduces 2 new api routes based on the https://llmstxt.org/ specification.
- llms.txt -> a directory of all markdown files + landing page
- llms-full.txt -> concatenation of all markdown pages

Included in this PR:
- titles and descriptions are extracted from all markdown page frontmatters, and stitched back together
- content of landing pages will always be included in `llms.txt`

Not included in this PR:
- no support for api endpoint markdown -> need way to represent api endpoint definition as markdown, with links to code examples.
- no support for markdown snippets or code snippets -> CLI/FDR needs to store more metadata about code snippets and we need to decide what the pathname is.

### DIFF
--- a/packages/ui/docs-bundle/src/middleware.ts
+++ b/packages/ui/docs-bundle/src/middleware.ts
@@ -63,6 +63,17 @@ export const middleware: NextMiddleware = async (request) => {
     }
 
     /**
+     * Rewrite llms-full.txt
+     */
+    if (pathname.endsWith("/llms-full.txt")) {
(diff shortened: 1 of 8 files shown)


dsinghvi (approved):
How do you handle versioned docs? do we create all the versions for the llm?

abvthecity:
@dsinghvi versioned docs are handled such that
- at the root level, all docs pages are joined together
- on the version level, only docs pages that are children of that version is joined together.

this works recursively too.

PR #1830, opened 2024-11-18
fix: llms.txt should show the overview page

abvthecity:
Similar to landing page, overview pages should be appended to the top of `llms.txt` and its children are enumerated below.


packages/ui/docs-bundle/src/pages/api/fern-docs/llms.txt.ts
```diff
@@ -85,7 +86,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
     FernNavigation.traverseDF(root, (node, parents) => {
         // don't include the landing page in the list
         if (landingPage != null && node.id === landingPage.id) {
-            return SKIP;
+            return CONTINUE;
```
abvthecity, line 89: this ensures that descendents are still visited

dsinghvi (approved):

PR #1836, opened 2024-11-19
Upgrade node version to node:20.18-alpine3.20

dubwub:


dubwub:
note: this can probably be closed, the vanta alert was cleared up

PR #1841, opened 2024-11-19
fix: changelog converter uses real year

RohinBhargava:

* fixes bug where changelog records were rendering with 1900

* some changelogs were rendering 1900 as the year in which they were created, this is not correct

* requires FDR update

### DIFF
--- a/packages/fdr-sdk/src/navigation/versions/v1/converters/ChangelogConverter.ts
+++ b/packages/fdr-sdk/src/navigation/versions/v1/converters/ChangelogConverter.ts
@@ -70,13 +70,7 @@ export class ChangelogNavigationConverter {
         entries: FernNavigation.V1.ChangelogEntryNode[],
         parentSlug: SlugGenerator,
     ): FernNavigation.V1.ChangelogYearNode[] {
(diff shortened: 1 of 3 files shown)


packages/fdr-sdk/src/navigation/versions/v1/converters/ChangelogConverter.ts
```diff
@@ -105,19 +105,21 @@ export class ChangelogNavigationConverter {
...
             months.set(month, monthEntries);
         }
         return orderBy(
-            Array.from(months.entries()).map(([month, entries]) =>
+            Array.from(months.entries()).map(([month, [year, entries]]) =>
                 this.#idgen.with(month.toString(), (id) => ({
                     id,
                     type: "changelogMonth" as const,
-                    title: dayjs(new Date(0, month - 1)).format("MMMM YYYY"),
+                    title: dayjs(new Date(year, month - 1)).format("MMMM YYYY"),
```
dsinghvi, line 122: up to add a unit test to confirm that behavior works as intended?

dsinghvi: specifically i could imagine pushing this into a utility function, testing that utility function and then using it here

PR #1846, opened 2024-11-21
feat: openapi parser array, null, enum, oneof, allof

RohinBhargava:

* Changed structure of nodes slightly to accommodate better ergonomics -- convert creates more natural fdr shapes, now
* Added Array, Null, Enum, and OneOf nodes
* Added allOf support for object nodes
* Added Deeptune snapshot test
* Would not expect this to be at full parity with the openapi parser, but is continuing to improve

* Fleshing out the OpenApi Schema Parser
* Will work on endpoint parser, availability and displayName extensions next

* Unit tests added for each new node,
* Added Deeptune snapshot test


packages/ui/fern-dashboard/src/routeTree.gen.ts
```diff
@@ -88,16 +91,81 @@ declare module '@tanstack/react-router' {
...
 // Create and export the route tree
 
-export const routeTree = rootRoute.addChildren({
-  IndexRoute,
-  AuthRoute: AuthRoute.addChildren({ AuthLoginRoute }),
-  AuthenticatedRoute: AuthenticatedRoute.addChildren({
-    AuthenticatedTeamOrgIdRoute,
-  }),
-})
+interface AuthRouteChildren {
```
dsinghvi, line 94: nit: lets remove this? or is this normal?

RohinBhargava: oh this is weird, I think this got added on compile, thanks for catching

abvthecity (approved):

packages/parsers/src/openapi/3.1/schemas/ArrayConverter.node.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export class ArrayConverterNode extends BaseOpenApiV3_1Node<
+    ArrayConverterNode.Input,
+    ArrayConverterNode.Output | undefined
+> {
+    innerSchema: SchemaConverterNode | undefined;
+
+    constructor(...args: BaseOpenApiV3_1NodeConstructorArgs<ArrayConverterNode.Input>) {
+        super(...args);
+        this.safeParse();
```
dsinghvi, line 24: curious -- why do we call `this.safeParse()` in the constructor? if we do this everywhere can we push this to `BaseOpenApiV3_1Node` ?

packages/parsers/src/openapi/3.1/schemas/BooleanConverter.node.ts
```diff
@@ -1,41 +1,48 @@
...
+    constructor(...args: BaseOpenApiV3_1NodeConstructorArgs<BooleanConverterNode.Input>) {
+        super(...args);
+        this.safeParse();
+    }
 
-        if (input.default != null && typeof input.default !== "boolean") {
+    parse(): void {
+        if (this.input.default != null && typeof this.input.default !== "boolean") {
             this.context.errors.warning({
                 message: "The default value for a boolean type should be a boolean",
```
dsinghvi, line 29: nit (unrelated to this PR) might be nice for the error message here `Expected example to be boolean. Received {}`

RohinBhargava: This is a good point, will fix all the error messages to accommodate this messaging.

packages/parsers/src/openapi/3.1/schemas/ArrayConverter.node.ts
RohinBhargava: This is an unfortunate part of TS. The issue is if there is a superclass call, the `this` doesn't get bound properly, and I believe the object members get associated with the proto of the super class (not 100% the cause, but 100% sure on the symptoms). The only real way to do this is to unfortunately include the constructor, and this is the slimmest I can make it.

packages/parsers/src/openapi/3.1/schemas/ArrayConverter.node.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    parse(): void {
+        this.innerSchema = new SchemaConverterNode(this.input.items, this.context, this.accessPath, "items");
+
+        if (this.input.items == null) {
+            this.context.errors.error({
+                message: "No items found in array",
```
dsinghvi, line 32: nit: `item` key missing in array schema

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -20,114 +24,109 @@ export class SchemaConverterNode extends BaseOpenApiV3_1Node<
...
-        pathId: string,
-    ) {
-        super(input, context, accessPath, pathId);
+    parse(): void {
+        // Check if the input is a reference object
+        if (isReferenceObject(this.input)) {
+            this.typeShapeNode = new ReferenceConverterNode(this.input, this.context, this.accessPath, this.pathId);
+        } else {
+            // If the object is not a reference object, then it is a schema object, gather all appropriate variables
+            this.name = this.input.title;
```
dsinghvi, line 43: what is the `name` used for?

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -20,114 +24,109 @@ export class SchemaConverterNode extends BaseOpenApiV3_1Node<
...
+                        case "null":
+                            this.typeShapeNode = new NullConverterNode(
+                                this.input as NullConverterNode.Input,
+                                this.context,
+                                this.accessPath,
+                                this.pathId,
+                            );
+                            break;
+                        default:
+                            new UnreachableCaseError(this.input.type);
```
dsinghvi, line 115: what does `UnreachableCaseError` do?

RohinBhargava: UnreachableCaseError is effectively the same as assertNever; it is used in more places in the platform packages, so decided to go with that utility

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
RohinBhargava: it's used in the Components -> Types conversion, and from OneOf schemas (for display names)

PR #1854, opened 2024-11-25
feat: OpenApi Parser x-fern-group-name and x-fern-availability extensions

RohinBhargava:

* adds an extension for parsing x-fern-availability
* adds an extension for parsing x-fern-group-name

* adds necessary extensions to the openapi parser

* unit tests


packages/parsers/src/openapi/3.1/extensions/availability-node.ts
```diff
@@ -0,0 +1,44 @@
...
+                this.availability = FdrAPI.Availability.InDevelopment;
+                break;
+            case "deprecated":
+                this.availability = FdrAPI.Availability.Deprecated;
+                break;
+            case undefined:
+                this.availability = undefined;
+                break;
+            default:
+                new UnreachableCaseError(inputWithAvailability["x-fern-availability"]);
```
dsinghvi, line 36: what would it look like if they had a value called "deep" for availability?

RohinBhargava: I imagine it would become one of the match arms, because UnreachableCaseError would give a compile break. On second thought this logic might be more appropriate in the convert function, and the parse would just set the availability as a member

dsinghvi (approved):

PR #1856, opened 2024-11-27
feat: OpenApi Parser parse requests and responses

RohinBhargava:

* parses request and response shapes on endpoints
* adds a couple nodes to discriminate on media/content types
* adds a parameterbase node for shared logic between many properties
* fixes minor papercuts with extensions

* rounding out endpoint parsing

* need to add unit tests


packages/parsers/src/openapi/3.1/schemas/OneOfConverter.node.ts
```diff
@@ -36,13 +29,13 @@ export class OneOfConverterNode extends BaseOpenApiV3_1ConverterNode<
...
-                            });
-                            return undefined;
-                        }
+                        // if (!isReferenceObject(schema) && schema.type !== "object") {
+                        //     this.context.errors.error({
+                        //         message: `Expected 'oneOf' schema to be an object. Received ${schema.type}`,
+                        //         path: this.accessPath,
+                        //     });
+                        //     return undefined;
+                        // }
```
dsinghvi, line 38: is the commented code here okay to delete? I don't think we'd want to bring it back

dsinghvi: and then we can delete the `.filter(isNonNullish)`

packages/parsers/src/openapi/types/format.types.ts
```diff
@@ -48,3 +48,22 @@ export const OPENAPI_STRING_TYPE_FORMAT = [
...
+export const SUPPORTED_REQUEST_CONTENT_TYPES = [
+    "application/json",
+    "multipart/form-data",
+    "application/octet-stream",
+] as const;
+export const SUPPORTED_RESPONSE_CONTENT_TYPES = [
+    "application/json",
+    "text/event-stream",
+    "application/octet-stream",
+] as const;
```
dsinghvi, line 67: this is a very restrictive set of content types -- can we support arbitrary content types? there are a lot of companies that do like `x-vnd-json` or `merge-patch-json`

packages/parsers/src/openapi/utils/3.1/convertToObjectProperties.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function convertToObjectProperties(
+    properties:
+        | Record<
+              string,
+              {
+                  availability?: AvailabilityConverterNode;
+                  description?: string;
+                  convert: () => FdrAPI.api.latest.TypeShape | undefined;
+              }
```
dsinghvi, line 13: nit: would be better to explicitly define and name this type so that someone can get more context on what this is

packages/parsers/src/openapi/utils/3.1/resolveReference.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    // Step 3: If the result is another reference object, make a recursive call
+    if (isReferenceObject(resolvedSchema)) {
+        resolvedSchema = resolveReference(resolvedSchema, document, defaultOutput);
+    }
+
+    // Step 4: If the result is a schema object, return it
+    return resolvedSchema;
+}
```
dsinghvi, line 33: nice! 🎉

packages/parsers/src/openapi/utils/3.1/resolveSchemaReference.ts
```diff
@@ -1,40 +1,13 @@
...
-    // Step 3: If the result is another reference object, make a recursive call
-    if (isReferenceObject(resolvedSchema)) {
-        resolvedSchema = resolveSchemaReference(resolvedSchema, document);
-    }
-
-    // Step 4: If the result is a schema object, return it
-    return resolvedSchema;
+    return resolveReference<OpenAPIV3_1.SchemaObject>(schema, document, {
+        "x-fern-type": "unknown",
+        additionalProperties: true,
```
dsinghvi, line 10: why are additionalProperties: true here? `unknown` implies that you can have whatever properties you want

packages/parsers/src/openapi/3.1/guards/isNonArraySchema.ts
```diff
@@ -0,0 +1,6 @@
+import { OpenAPIV3_1 } from "openapi-types";
+import { isArraySchema } from "./isArraySchema";
+
+export function isNonArraySchema(input: OpenAPIV3_1.SchemaObject): input is OpenAPIV3_1.NonArraySchemaObject {
+    return !Array.isArray(input.type) && !isArraySchema(input);
```
dsinghvi, line 5: what is a `nonArraySchema` ?

dsinghvi: when we talk about an array schema, which are we talking about: 

```yml
type: [ 'string', 'null' ] 
```
or 
```yml
type: array
items: 
  schema: ...
```

packages/parsers/src/openapi/3.1/guards/isObjectSchema.ts
```diff
@@ -2,5 +2,5 @@ import { OpenAPIV3_1 } from "openapi-types";
 import { ObjectConverterNode } from "../schemas/ObjectConverter.node";
 
 export function isObjectSchema(input: OpenAPIV3_1.SchemaObject): input is ObjectConverterNode.Input {
-    return typeof input.type === "string" && input.type === "object";
+    return input.type == null || (typeof input.type === "string" && input.type === "object");
```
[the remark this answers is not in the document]
dsinghvi: i think you want to check `type == null` and `properties` are present but `oneOf` and `anyOf` are undefined?

packages/parsers/src/openapi/3.1/paths/parameters/ParameterBaseObjectConverter.node.ts
```diff
@@ -0,0 +1,72 @@
...
+import { OpenAPIV3_1 } from "openapi-types";
+import {
+    BaseOpenApiV3_1ConverterNode,
+    BaseOpenApiV3_1ConverterNodeConstructorArgs,
+} from "../../../BaseOpenApiV3_1Converter.node";
+import { AvailabilityConverterNode } from "../../extensions/AvailabilityConverter.node";
+import { isReferenceObject } from "../../guards/isReferenceObject";
+import { SchemaConverterNode } from "../../schemas/SchemaConverter.node";
+
+export class ParameterBaseObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
```
dsinghvi, line 11: nit: lets call this `ParameterConverterNode` ?

packages/parsers/src/openapi/3.1/paths/parameters/ParameterBaseObjectConverter.node.ts
```diff
@@ -0,0 +1,72 @@
...
+                    context: this.context,
+                    accessPath: this.accessPath,
+                    pathId: "availability",
+                });
+                this.schema = new SchemaConverterNode({
+                    input: this.input.schema,
+                    context: this.context,
+                    accessPath: this.accessPath,
+                    pathId: "schema",
+                });
```
dsinghvi, line 58:
```ts
this.schema = new SchemaConverterNode({
                    input: this.input.schema,
                    context: this.context,
                    accessPath: this.accessPath,
                    pathId: "schema",
                });
```
nit: can we refactor this out and remove any sort of duplication like this?

dsinghvi: for both `IsReferenceObject` and schema

packages/parsers/src/openapi/3.1/paths/parameters/ParameterBaseObjectConverter.node.ts
```diff
@@ -0,0 +1,72 @@
...
+                pathId: "schema",
+            });
+            this.availability = new AvailabilityConverterNode({
+                input: this.input,
+                context: this.context,
+                accessPath: this.accessPath,
+                pathId: "availability",
+            });
+        } else {
+            if (this.input.schema != null) {
```
dsinghvi, line 46: parameter schemas may often be null, when they are i think it makes sense to model it as a optional string

packages/parsers/src/openapi/3.1/paths/parameters/ParameterBaseObjectConverter.node.ts
```diff
@@ -0,0 +1,72 @@
...
+                this.schema = new SchemaConverterNode({
+                    input: this.input.schema,
+                    context: this.context,
+                    accessPath: this.accessPath,
+                    pathId: "schema",
+                });
+                this.required = this.input.required;
+            } else {
+                this.context.errors.error({
+                    message: `Expected reference or schema for parameter. Received: ${JSON.stringify(this.input)}`,
```
dsinghvi, line 62: lets never do the `JSON.stringify()` its pretty disgusting and a mistake that we ever added to the CLI. https://github.com/fern-api/fern-platform/pull/1856/files#r1866208859 I do think we can omit this log line and just treat as optional string though as mentioned in this comment

packages/parsers/src/openapi/3.1/paths/request/RequestMediaTypeObjectConverter.node.ts
```diff
@@ -0,0 +1,178 @@
...
+                this.schema = new ReferenceConverterNode({
+                    input: this.input.schema,
+                    context: this.context,
+                    accessPath: this.accessPath,
+                    pathId: "schema",
+                });
+            } else if (isObjectSchema(this.input.schema)) {
+                const mediaType = MediaType.parse(contentType);
+                // An exhaustive switch cannot be used here, because contentType is an unbounded string
+                if (mediaType?.isJSON()) {
```
dsinghvi, line 58: nit: the thing that has been successful in the previous OpenAPI converter is if the content type contains json. It may be more clear to rename this `mediaType.containsJson()` ?

packages/parsers/src/openapi/3.1/paths/request/RequestMediaTypeObjectConverter.node.ts
```diff
@@ -0,0 +1,178 @@
...
+                    this.contentType = "application/json" as const;
```
dsinghvi, line 59: we should use the content type directly from the schema so we handle things like `application/x-vnd-json`

packages/parsers/src/openapi/3.1/paths/request/MultipartFormDataPropertySchemaConverter.node.ts
```diff
@@ -0,0 +1,49 @@
...
+        super(args);
+        this.safeParse();
+    }
+
+    override parse(): void {
+        super.parse();
+        const propertyObject = isReferenceObject(this.input)
+            ? resolveSchemaReference(this.input, this.context.document)
+            : this.input;
+        if (propertyObject.type === "string" && propertyObject.format === "binary") {
```
dsinghvi, line 26: nit: can we move this into a helper

packages/parsers/src/openapi/3.1/paths/request/MultipartFormDataPropertySchemaConverter.node.ts
```diff
@@ -0,0 +1,49 @@
...
+            this.multipartType = "file";
+            this.contentType = propertyObject.contentMediaType;
+        } else if (propertyObject.type === "array") {
+            const items = isReferenceObject(propertyObject.items)
+                ? resolveSchemaReference(propertyObject.items, this.context.document)
+                : propertyObject.items;
+            if (items?.type === "string" && items?.format === "binary") {
```
dsinghvi, line 33: nit: reuse helper from above

packages/parsers/src/openapi/3.1/paths/request/MultipartFormDataPropertySchemaConverter.node.ts
```diff
@@ -0,0 +1,49 @@
...
+import { OpenAPIV3_1 } from "openapi-types";
+import { BaseOpenApiV3_1ConverterNodeConstructorArgs } from "../../../BaseOpenApiV3_1Converter.node";
+import { ConstArrayToType, SUPPORTED_MULTIPART_TYPES } from "../../../types/format.types";
+import { resolveSchemaReference } from "../../../utils/3.1/resolveSchemaReference";
+import { isReferenceObject } from "../../guards/isReferenceObject";
+import { SchemaConverterNode } from "../../schemas/SchemaConverter.node";
+
+export type MultipartType = ConstArrayToType<typeof SUPPORTED_MULTIPART_TYPES>;
+
+export class MultipartFormDataPropertySchemaConverterNode extends SchemaConverterNode {
```
dsinghvi, line 12: this is cool re extending `SchemaConverterNode` 🎉 

Any downsides that are not immediately obvious?

packages/parsers/src/openapi/3.1/paths/response/ResponsesObjectConverter.node.ts
```diff
@@ -0,0 +1,117 @@
...
+import { convertToObjectProperties } from "../../../utils/3.1/convertToObjectProperties";
+import { ResponseObjectConverterNode } from "./ResponseObjectConverter.node";
+
+export class ResponsesObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
+    OpenAPIV3_1.ResponsesObject,
+    {
+        responses: {
+            headers: FdrAPI.api.latest.ObjectProperty[] | undefined;
+            response: FdrAPI.api.latest.HttpResponse;
+        }[];
```
dsinghvi, line 17: does FDR support this sort of model? (or do you have to add support eventually?)

packages/parsers/src/openapi/3.1/paths/OperationObjectConverter.node.ts
```diff
@@ -113,26 +150,28 @@ export class OperationObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
...
-            request: undefined,
-            response: undefined,
-            errors: [],
+            pathParameters: convertToObjectProperties(this.pathParameters),
+            queryParameters: convertToObjectProperties(this.queryParameters),
+            requestHeaders: convertToObjectProperties(this.requestHeaders),
+            responseHeaders: responses?.[0]?.headers,
+            // TODO: revisit fdr shape to suport multiple requests
+            request: this.requests?.convert()[0],
+            response: responses?.[0]?.response,
```
dsinghvi, line 173: nit: can you add a debug log whenever we index into [0] ?

dsinghvi: this is across the whole parser

packages/parsers/src/openapi/3.1/guards/isObjectSchema.ts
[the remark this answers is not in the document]
RohinBhargava: yeah, so that second check is deferred to the actual validation of the node -- we can have a stricter check in here, but I think it might be more effective to leverage the error collector -- the guard is just to satisfy the type checker.

packages/parsers/src/openapi/3.1/paths/parameters/ParameterBaseObjectConverter.node.ts
RohinBhargava: this is mapping from the openapi-types name, not sure if we want to do that?

packages/parsers/src/openapi/3.1/paths/request/RequestMediaTypeObjectConverter.node.ts
RohinBhargava: sounds good, I was treating this like a discriminant to enumerate all possible cases, but I'll think about a better discriminant!

packages/parsers/src/openapi/3.1/paths/response/ResponsesObjectConverter.node.ts
RohinBhargava: we'll have to add support -- right now, there's only one response in FDR, so the request/response section needs some schematic reworking

packages/parsers/src/openapi/3.1/guards/isNonArraySchema.ts
RohinBhargava: it is both. The former is an example of a "MixedSchemaObject" type and the latter is an example of an "ArraySchemaObject" in OpenApi. We only want the NonArraySchemaObject from this guard, which has a discrete type

RohinBhargava:
```
    export type SchemaObject = ArraySchemaObject | NonArraySchemaObject | MixedSchemaObject;
    export interface ArraySchemaObject extends BaseSchemaObject {
        type: ArraySchemaObjectType;
        items: ReferenceObject | SchemaObject;
    }
    export interface NonArraySchemaObject extends BaseSchemaObject {
        type?: NonArraySchemaObjectType;
    }
    interface MixedSchemaObject extends BaseSchemaObject {
        type?: (ArraySchemaObjectType | NonArraySchemaObjectType)[];
        items?: ReferenceObject | SchemaObject;
    }
```

If we want to handle the MixedSchemaObject, I think this should return a list of SchemaObjectConverterNodes, that are handled one by one. ArraySchemaObjects are already handled

packages/parsers/src/openapi/3.1/paths/OperationObjectConverter.node.ts
```diff
@@ -33,21 +39,32 @@ export class OperationObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
...
-                //     pathId: `parameters[${index}]`,
-                // });
-            } else if (parameter.in === "path") {
+                const resolvedParameter = resolveParameterReference(parameter, this.context.document);
+                if (resolvedParameter != null) {
+                    parameter = resolvedParameter;
+                } else {
+                    this.context.errors.warning({
+                        message: "Expected parameter reference to resolve to a parameter object. Received null",
+                        path: this.accessPath,
```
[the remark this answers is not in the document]
RohinBhargava: This would be the path of the object, followed by one of "get", "post",  "put", "delete". I don't think OpenApi supports a list of APIs below a path and method

packages/parsers/src/openapi/3.1/paths/request/MultipartFormDataPropertySchemaConverter.node.ts
RohinBhargava: None that I can think of, other than it might be a little untype-safe to use in certain places -- can't constrain an input. If this is the case, I think we can add an interface that can be passed into constructor args, so I don't see many downsides

PR #1857, opened 2024-11-28
Rohin/openapi parser publishing indexes

RohinBhargava:

* add indexes to project to export nicely
* change to fdr-cjs-sdk for branded aliases

* hooking the project up to the CLI

* manually tested with cohere openapi and dev cli

### DIFF
--- a/packages/parsers/package.json
+++ b/packages/parsers/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@fern-api/docs-parsers",
-  "version": "0.0.2",
+  "version": "0.0.3",
   "repository": {
     "type": "git",
     "url": "https://github.com/fern-api/fern-platform.git",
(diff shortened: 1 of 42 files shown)


packages/parsers/package.json
```diff
@@ -37,5 +41,8 @@
     "@types/uuid": "^9.0.1",
     "@types/whatwg-mimetype": "^3.0.2",
     "js-yaml": "^4.1.0"
+  },
+  "peerDependencies": {
+    "@fern-fern/fdr-cjs-sdk": "0.116.5-873e41db0"
```
dsinghvi, line 46: pinning this seems like a mistake still -- can we generate the types into here?

dsinghvi: lets add a third block here https://github.com/fern-api/fern-platform/blob/1fce154cd39922c8866ba3f981e11f83580db9bb/fern/apis/fdr/generators.yml#L16-L28

PR #1859, opened 2024-12-04
fix: truncate to 10kb for mardown sections

RohinBhargava:

* truncates markdown-section-v1 records to a max size of 50kb

* customer generation failures

* to be tested in fdr dev


dsinghvi (approved):

packages/ui/fern-docs-search-server/src/algolia/records/generateEndpointRecords.ts
```diff
@@ -16,17 +17,18 @@ export function generateEndpointRecord({
...
         method: endpoint.method,
         endpointPath: endpoint.path,
         isResponseStream: endpoint.response?.body.type === "stream" || endpoint.response?.body.type === "streamingText",
         title: node.title,
-        description: toDescription([
-            endpoint.description,
-            endpoint.request?.description,
-            endpoint.response?.description,
-        ]),
+        description: description?.length ? truncateToBytes(description, 10 * 1024) : undefined,
```
RohinBhargava, line 31: Will a collection of strings here produce multiple records? I'm probably mistaken, but I think we want to use the function outside and then push the records with each one of the truncated descriptions. Something like:
```
truncateToBytes(description, ...).map((des) => records.push({
...
description: des
...
}`?

PR #1882, opened 2024-12-10
fix(docs): Sidebar navigation for anchors

sbelkadi:
Fixes FER-3720

The sidebar now highlights anchor nodes that are currently selected. The mobile sidebar now closes when an anchor node is selected.

Anchor nodes in sidebars were not behaving the same as section nodes and page nodes:
- Anchor nodes would not remain highlighted in the sidebar when selected
- Anchor nodes would not close the mobile sidebar when selected

Tested locally on a section sidebar. Tested with both the desktop and mobile sidebars.


packages/ui/app/src/sidebar/SidebarLink.tsx
```diff
@@ -86,6 +86,16 @@ const SidebarLinkInternal = forwardRef<HTMLDivElement, SidebarLinkProps>((props,
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     useImperativeHandle(parentRef, () => ref.current!);
 
+    const closeMobileSidebar = useCloseMobileSidebar();
+
+    const handleClick = useCallback<React.MouseEventHandler<HTMLAnchorElement | HTMLButtonElement>>(
+        (e) => {
+            onClick?.(e);
+            closeMobileSidebar();
```
abvthecity, line 94: toggleExpand got dropped here. I think if toggleExpand is present, then you don't close the mobile sidebar

sbelkadi: Chatted live with Andrew - toggleExpand wasn't closing the sidebar, so it might be farther upstream. For now, closeMobileSidebar doesn't seem to affect other sites.

abvthecity (approved):

PR #1894, opened 2024-12-13
feat: OpenApi Parser Examples and Subpackage ergonomics

RohinBhargava:

* parses examples,
* adds extensions,
* refactors extension ergonomics

* continuation of parser

* With realtime tests


packages/parsers/src/openapi/3.1/schemas/ReferenceConverter.node.ts
```diff
@@ -37,7 +37,7 @@ export class ReferenceConverterNode extends BaseOpenApiV3_1ConverterNode<
             type: "alias",
             value: {
                 type: "id",
-                id: FernRegistry.TypeId(this.schemaId),
+                id: FernRegistry.TypeId(`type_:${this.schemaId}`),
```
dsinghvi, line 40: nit: @RohinBhargava @abvthecity thoughts on just making this the schemaId directly? this will make it easier to eventually build a `<Schema>` component because we'll just have to match against the id instead of doing any sort of string manipulation

RohinBhargava: yup, I thought I needed to add this for some ui, but I'll remove and re-add if that's the case

dsinghvi (approved):
+1ing so that you can get these in -- but chatted offline but moving to named params + reducing nesting by using helper functions

RohinBhargava:
Added an item to the tracking notion to improve hygeine

PR #1901, opened 2024-12-16
chore: modify fdr apis to accomodate openapi parser for docs generation

RohinBhargava:

* Modifies FDR apis to enable `docs dev` to run with the new openapi parser

* in order for the docs bundle to pick up the correct shape from FDR, we need to mock the endpoints. This modified FDR and the docs 
* In addition, it makes the apis functional with Prisma updates as well, which should also enable docs generation

* To be tested in fdr dev with cli


packages/ui/fern-docs-server/src/ApiDefinitionLoader.ts
```diff
@@ -101,17 +101,26 @@ export class ApiDefinitionLoader {
...
                 throw new Error("Failed to load API definition");
             }
         }
 
-        return ApiDefinitionV1ToLatest.from(v1.body, this.flags).migrate();
+        return v1.ok
+            ? ApiDefinitionV1ToLatest.from(v1.body, this.flags).migrate()
+            : latest.ok
+              ? latest.body
+              : undefined;
```
dsinghvi, line 123: nit this code might read better if you try to get v1 and then return if it succeeds and then if you try to v2 and then return if it succeeds

dsinghvi (approved):
looks great -- probably worth heavily testing on dev

fern/apis/fdr/definition/api/latest/__package__.yml
```diff
@@ -9,6 +9,21 @@ imports:
   rootCommons: ../../commons.yml
   auth: auth.yml
 
+service:
+  base-path: /registry/api/latest
```
abvthecity, line 13: This needs to be more discretely versioned because "latest" can change in the future.

fern/apis/fdr/definition/api/v1/register/__package__.yml
```diff
@@ -23,7 +24,8 @@ service:
           properties:
             orgId: rootCommons.OrgId
             apiId: rootCommons.ApiId
-            definition: ApiDefinition
+            definition: optional<ApiDefinition>
+            definitionLatest: optional<latest.ApiDefinition>
```
abvthecity, line 28: Also needs to be discretely versioned

fern/apis/fdr/definition/docs/v1/read/__package__.yml
```diff
@@ -67,6 +68,7 @@ types:
       algoliaSearchIndex: optional<algolia.AlgoliaSearchIndex>
       pages: map<rootCommons.PageId, PageContent>
       apis: map<rootCommons.ApiDefinitionId, apiReadV1.ApiDefinition>
+      apisLatest: map<rootCommons.ApiDefinitionId, apiReadLatest.ApiDefinition>
```
abvthecity, line 71: Here too

packages/parsers/.gitignore
```diff
@@ -1,6 +1,7 @@
 openapi/shared/temporary
 src/client/generated/Client.ts
 src/client/generated/api/resources/api/client/
+src/client/generated/api/resources/api/resources/latest/client/
```
abvthecity, line 4: This should be globbed generated/*

packages/parsers/tsconfig.json
```diff
@@ -12,6 +12,7 @@
     "node_modules",
     "src/client/generated/Client.ts",
     "src/client/generated/api/resources/api/client/",
+    "src/client/generated/api/resources/api/resources/latest/client/",
```
abvthecity, line 15: Nit: Globbed here too

packages/ui/app/src/resolver/resolveDocsContent.ts
```diff
@@ -64,14 +66,25 @@ export async function resolveDocsContent({
...
-        return ApiDefinitionLoader.create(domain, api.id)
-            .withMdxBundler(serializeMdx, engine)
-            .withFlags(featureFlags)
-            .withApiDefinition(ApiDefinitionV1ToLatest.from(api, featureFlags).migrate())
-            .withEnvironment(process.env.NEXT_PUBLIC_FDR_ORIGIN)
-            .withResolveDescriptions();
-    });
+    // TODO: remove legacy when done
+    const apiLoaders =
+        Object.keys(apis).length > 0
```
abvthecity, line 71: This should be rewritten such that the keys are merged together, and the callback should try latest and fallback to legacy for a more graceful transition from legacy -> latest

packages/ui/fern-docs-search-server/src/fdr/load-docs-with-url.ts
```diff
@@ -63,13 +63,16 @@ export async function loadDocsWithUrl(payload: LoadDocsWithUrlPayload): Promise<
...
-    const apis = mapValues(docs.body.definition.apis, (api) =>
-        ApiDefinition.ApiDefinitionV1ToLatest.from(api, {
-            useJavaScriptAsTypeScript: payload.useJavaScriptAsTypeScript ?? false,
-            alwaysEnableJavaScriptFetch: payload.alwaysEnableJavaScriptFetch ?? false,
-            usesApplicationJsonInFormDataValue: payload.usesApplicationJsonInFormDataValue ?? false,
-        }).migrate(),
-    );
+    const apis =
+        Object.keys(docs.body.definition.apis).length > 0
+            ? mapValues(docs.body.definition.apis, (api) =>
```
abvthecity, line 68: Should be merged more gracefully

packages/ui/fern-docs-server/src/ApiDefinitionLoader.ts
```diff
@@ -106,12 +106,30 @@ export class ApiDefinitionLoader {
...
-                console.error(v1.error.content);
+                console.error(v1?.error?.content);
                 throw new Error("Failed to load API definition");
             }
         }
+        if (v1.ok) {
+            return ApiDefinitionV1ToLatest.from(v1.body, this.flags).migrate();
+        }
+
+        const latest = await this.#getClient().api.latest.getApiLatest(this.apiDefinitionId);
```
abvthecity, line 117: Should reverse this logic so that the v1 is a fallback.

servers/fdr/src/controllers/api/getRegisterApiService.ts
```diff
@@ -21,63 +21,80 @@ export function getRegisterApiService(app: FdrApplication): APIV1WriteService {
...
-            }
-            if (snippetsConfigurationWithSdkIds.goSdk != null) {
-                sdkIds.push(snippetsConfigurationWithSdkIds.goSdk.sdkId);
-            }
-            if (snippetsConfigurationWithSdkIds.rubySdk != null) {
-                sdkIds.push(snippetsConfigurationWithSdkIds.rubySdk.sdkId);
+            const apiDefinitionId = FdrAPI.ApiDefinitionId(uuidv4());
+            let transformedApiDefinition: APIV1Db.DbApiDefinition | FdrAPI.api.latest.ApiDefinition | undefined;
+
+            if (req.body.definition != null) {
```
abvthecity, line 28: This seems to drop snippet configuration from latest?

fern/apis/fdr/definition/api/latest/__package__.yml
RohinBhargava: I think this api should return what latest is at any given time, which should not be tied to a specific ApiDefinition, but point understood on the other fields

packages/parsers/.gitignore
RohinBhargava: not everything under generated is ignored, only specific pieces not relevant to types. I can try src/client/generated/**/client/,
src/client/generated/**/Client.ts

servers/fdr/src/controllers/api/getRegisterApiService.ts
RohinBhargava: there is no snippet configuration in latest definition anyway. It is out of scope for right at this very moment, since there are V1 specific converters that can be solved with @dsinghvi's example generator library

PR #1940, opened 2024-12-23
fix(docs): fix feedback form logic

chdeskur:
This updates the logic for the feedback form to ensure the correct responses are displayed.


eyw520 (approved):
makes sense to me

PR #1944, opened 2024-12-24
chore(deps): upgrade fontawesome

chdeskur:
This PR upgrades the version of Font Awesome. It also updates the logic of our Font Awesome Icon Parser to accommodate icons that include multiple styles (e.g. sharp thin house)

Before:
Screenshot 2024-12-24 at 3 28 21 PM

After:
Screenshot 2024-12-24 at 3 28 13 PM


packages/fern-docs/components/src/FontAwesomeIcon.tsx
```diff
@@ -67,12 +67,19 @@ function getCdnHost() {
...
+    return ["solid", parts[0]];
+  }
+
+  const iconName = parts.pop();
+  if (iconName == null) {
+    return;
   }
-  return;
+  const style = parts.join("-");
+  return [style, iconName];
```
dsinghvi, line 84: @chdeskur do you mind adding some brief comments about what this function is doing? 

My understanding: if the icon string has no style, it defaults to `fa-solid {icon}`. Not sure if i follow what comes after? 

What is this part doing: 
```ts
const style = parts.join("-");
return [style, iconName];
```

dsinghvi (approved):

PR #1946, opened 2024-12-30
Adding filesize/mime-type optional fields to FDR presigned URL generation

dubwub:


servers/fdr/src/services/s3/S3Service.ts
```diff
@@ -205,30 +259,45 @@ export class S3ServiceImpl implements S3Service {
...
-    const input: PutObjectCommandInput = {
+    const key = this.constructS3DocsKey({ domain, time, filepath });
+
+    let conditions: any[] | undefined = [];
+    if (typeof filesize !== "undefined") {
+      conditions.push(["content-length-range", filesize, filesize]);
+    }
+    if (typeof mimeType !== "undefined") {
+      if (!ALLOWED_FILE_TYPES.has(mimeType)) {
+        throw new Error("Invalid mime-type: " + mimeType); // TODO: are these generated somewhere?
```
dsinghvi, line 284: I would create a new error in the fern definition for `fdr` that takes a list of flies that are too large and a list of flies that have an unallowed mimetype

dsinghvi: then you can collect all the issues and throw that error once with all the info

dsinghvi: The CLI can then read that error and log a more sensible output for the user

dubwub: * new error added
* errors are now caught and collected and then re-thrown as a group

PR #1965, opened 2025-01-08
fix: FDR generated code imports

RohinBhargava:

* Adds the explicit types that generated TS code relies on (`@types/express-serve-static-core@0.4.19`) that was causing fdr to not compile
* Removes excluding compiling generated code that is necessary for FDR to run (https://github.com/fern-api/fern-platform/blob/6506692288b1102cea252a6fd3cc99afffe5b18a/servers/fdr/tsconfig.json)
* Updates pnpm lock

* FDR deployments have been broken for some time

* To be tested

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2750,6 +2750,9 @@ importers:
       '@types/express':
         specifier: ^4.17.13
         version: 4.17.21
+      '@types/express-serve-static-core':
+        specifier: 4.19.0
+        version: 4.19.0
       '@types/html-to-text':
         specifier: ^9.0.1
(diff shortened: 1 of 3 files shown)


rpc333 (approved):
LGTM!

Note for future readers: the `tsconfig` was changed as part of #1917 to exclude `**/generated` which then caused the issues here. That change was probably just an incidental thing to get the large PR to compile properly, but introduced this undesired side effect in FDR. The fix here is very sharp

https://github.com/fern-api/fern-platform/blob/6506692288b1102cea252a6fd3cc99afffe5b18a/servers/fdr/tsconfig.json

PR #1972, opened 2025-01-09
feat: expose searchbar as a component

chdeskur:
initial work to expose the searchbar component for general use


packages/fern-docs/ui/src/atoms/search.ts
```diff
@@ -3,6 +3,7 @@ import { atom, useAtom } from "jotai";
 export type ValidSegmentationKeys = "endpoints" | "pages" | "fields";
 
 const SEARCH_SEGMENTATION_ATOM = atom<ValidSegmentationKeys>("endpoints");
+export const PAGE_HAS_SEARCHBAR_ATOM = atom<boolean>(false);
```
abvthecity, line 6: what is this for? and it doesn't seem to derive state or is set by anything?

chdeskur: whoops yep, need to finish setting that up

abvthecity (approved):

PR #1991, opened 2025-01-12
fix(ai): only index nodes with unique canonical slugs

dsinghvi:
Node IDs are getting duplicated (root cause is unclear so far) 

Unblock Turbopuffer indexing at a customer

Yes, locally tested

### DIFF
--- a/packages/fdr-sdk/src/navigation/NodeCollector.ts
+++ b/packages/fdr-sdk/src/navigation/NodeCollector.ts
@@ -241,7 +241,7 @@ export class NodeCollector {
 
     [...this.slugToNode.values()]
       .filter(({ node }) => FernNavigation.isPage(node))
-      .filter(({ node }) => !node.hidden && !node.authed)
+      .filter(({ node }) => !node.hidden)
       .filter(({ node }) =>
         FernNavigation.hasMarkdown(node) ? !node.noindex : true
       )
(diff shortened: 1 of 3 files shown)


RohinBhargava (approved):

packages/fern-docs/search-server/src/turbopuffer/records/create-base-record.ts
```diff
@@ -71,7 +71,7 @@ export function createBaseRecord({
   );
 
   return {
-    id: `${org_id}:${domain}:${node.id}`,
+    id: `${org_id}:${domain}:${node.slug}`,
```
abvthecity, line 74: id is more unique than slug

dsinghvi: I wrote a test locally and for some reason its not behaving that way

packages/fdr-sdk/src/navigation/NodeCollector.ts
```diff
@@ -236,6 +236,29 @@ export class NodeCollector {
     return this.#getIndexablePageSlugs();
   }
 
+  #getIndexablePageSlugsWithAuth = once((): NavigationNodeWithMetadata[] => {
```
abvthecity, line 239: getIndexabalePageNodesWithAuth

abvthecity (approved):

packages/fdr-sdk/src/navigation/NodeCollector.ts
dsinghvi: makes sense

PR #1994, opened 2025-01-12
feat: FDR Latest pulling at the edge (S3 upload/download)

RohinBhargava:

* This change introduces pulling FDR Latest specs directly from S3, using already existing S3 asset buckets. The reason we can do this is because we no longer have shape transformations for FDR docs shapes, and instead have one shape that is registered and read.

* This PR chooses to use FDR to upload S3 specs, however, this can easily be moved into the CLI (before this, there should be a discussion of security concerns of distributing S3 keys to clients/this is not an issue in Next since the api keys are used in the backend part of next). For spec pulling, this is done directly at the site that needs the data (some FDR apis or edge functions, because we need to index algolia records).

*The idea is that this should improve performance, since we see that FDR is a current bottleneck, and that there is one less hop to deliver the data to the end user.

*THIS NEEDS TO BE THOROUGHLY TESTED*

* Coupling this with the OpenApi Parser v2 rollout is beneficial, because we can properly flight/test this new approach.

* Performance issues
* Future vision to parse data at the edge
* Integration with dynamic snippets

* To be tested in FDR dev.


dsinghvi (approved):
It's a large PR, but this makes sense to me

PR #1996, opened 2025-01-13
feat: Record and send audio via Chrome microphone

namanagar:
Fixes [FER-3984](https://linear.app/buildwithfern/issue/FER-3984/[hume-elevenlabs-deepgram]-record-and-send-audio-via-chrome-microphone)

Adds hook for audio recording logic + UI to handle both base64 raw audio & audio file upload cases

Makes playground more intuitive for AI audio use cases

## Demo

https://github.com/user-attachments/assets/be0bc05b-1954-4449-958d-c19a1a865aad

## Caveats

- has not been tested against deeply nested base64 audio inputs (probably not an issue...?)
- currently only records/sends webm audio


packages/fern-docs/ui/src/playground/form/PlaygroundTypeReferenceForm.tsx
```diff
@@ -145,14 +146,27 @@ export const PlaygroundTypeReferenceForm =
...
               {hasVoiceIdPlaygroundForm && property?.key === "voice_id" ? (
-                // TODO: delete this:
                 <PlaygroundElevenLabsVoiceIdForm
                   id={id}
                   className="w-full"
                   value={typeof value === "string" ? value : ""}
                   onValueChange={onChange}
                   disabled={disabled}
                 />
+              ) : property?.key === "user_audio_chunk" || // user_audio_chunk hardcoded for ElevenLabs use case
```
RohinBhargava, line 156: Can we add a `TODO(naman)` here to remove this once the value is plumbed correctly

packages/fern-docs/ui/package.json
```diff
@@ -122,6 +122,7 @@
     "unist-util-visit": "^5.0.0",
     "url-join": "5.0.0",
     "use-memo-one": "^1.1.3",
+    "webm-duration-fix": "^1.0.4",
```
namanagar, line 125: tiny package that helps w/ the blob data from the media recorder api — without this, the `<audio>` element after recording has incorrect metadata (duration: Infinity) which messes with the playback

RohinBhargava (commented):
The FDR changes look great to me! Will leave to @abvthecity to approve for all of the UI related elements

packages/parsers/src/client/generated/api/resources/api/resources/v1/resources/register/resources/type/types/StringType.ts
```diff
@@ -8,4 +8,5 @@ export interface StringType {
     minLength: number | undefined;
     maxLength: number | undefined;
     default: string | undefined;
+    mimeType: string | undefined;
```
abvthecity, line 11: hot take: i'd probably want to remove this from the `string` type since the `base64` type already implies that it's data.

packages/parsers/src/client/generated/api/resources/api/resources/v1/resources/register/resources/type/types/Base64Type.ts
```diff
@@ -4,4 +4,5 @@
 
 export interface Base64Type {
     default: string | undefined;
+    mimeType: string | undefined;
```
abvthecity, line 7: i forgot about the base64 type; this is probably a good candidate for rendering a file dropdown for all base64-type string fields (along with the option to paste in a base64 string).

packages/fern-docs/ui/src/playground/form/PlaygroundWaveformAnimation.tsx
```diff
@@ -0,0 +1,22 @@
+import { motion } from "framer-motion";
+
+export function WaveformAnimation({ volume }: { volume: number }) {
+  return (
+    <div className="flex size-full items-center justify-between gap-0.5">
+      {Array.from({ length: 20 }).map((_, i) => (
```
abvthecity, line 6: nit: it would be good to include any todo comments

packages/fern-docs/ui/src/playground/hooks/useAudioRecorder.ts
```diff
@@ -0,0 +1,138 @@
...
+  const chunksRef = useRef<Blob[]>([]);
+  const [audioUrl, setAudioUrl] = useState<string | null>(null);
+  const mimeType = "audio/webm;codecs=opus";
+
+  const startRecording = useCallback(async () => {
+    try {
+      setElapsedTime(0);
+      chunksRef.current = [];
+
+      const stream = await navigator.mediaDevices.getUserMedia({ audio: true });
```
abvthecity, line 36: nit: it would be good to check if the mimetype is supported by the mediadevices, so we can hide this UX if not supported (this is probably relevant for iOS safari)

packages/fern-docs/ui/src/playground/form/PlaygroundTypeReferenceForm.tsx
```diff
@@ -153,6 +154,18 @@ export const PlaygroundTypeReferenceForm =
                   onValueChange={onChange}
                   disabled={disabled}
                 />
+              ) : property?.key === "user_audio_chunk" || // TODO(naman): remove hardcoding for ElevenLabs once the backend mimeType is plumbed through
+                (primitive.value.type === "base64" &&
+                  primitive.value.mimeType?.startsWith("audio/")) ? (
```
abvthecity, line 159: mimeType?.includes("audio/webm")

i think it's important to ensure that the customer who configures this microphone experience is _explicit_ that their server can support `audio/webm` otherwise we don't show the microphone experience at all

packages/fern-docs/ui/src/playground/form/PlaygroundFileUploadForm.tsx
```diff
@@ -104,11 +118,12 @@ export const PlaygroundFileUploadForm = memo<PlaygroundFileUploadFormProps>(
       >
         <input
           ref={ref}
-          type="file"
           id={id}
-          onChange={handleChange}
-          className="hidden"
+          type="file"
+          accept="audio/*"
```
abvthecity, line 123: we should pass in the mimetypes here if it's present in the api definition spec. otherwise, accept all filetypes.

abvthecity (approved):

packages/fern-docs/ui/src/playground/form/PlaygroundFileUploadForm.tsx
namanagar: for now accepting all filetypes & added a todo to be addressed in the followup PR: https://github.com/fern-api/fern-platform/pull/2012

PR #1999, opened 2025-01-13
feat: queue up nav config

chdeskur:
This PR adds the shape to include a configuration to set `hide-nav-links: true` globally in `docs.yml`


RohinBhargava (approved):
Looks good -- looks like you may have to plumb the values with undefined in places. I would also maybe consider a different PR title? Not sure how it relates

dsinghvi:
@chdeskur you'll need to update here https://github.com/fern-api/fern-platform/blob/b6ae97a67f9d50093cdfd03b5526cdee103e0fb3/packages/fdr-sdk/src/converters/db/convertDocsDefinitionToDb.ts#L95-L127 and https://github.com/fern-api/fern-platform/blob/b6ae97a67f9d50093cdfd03b5526cdee103e0fb3/packages/fdr-sdk/src/converters/read/convertDbDocsConfigToRead.ts#L21-L45

PR #2011, opened 2025-01-14
fix: OpenApi Parser v2: Strings, Webhooks, Objects

RohinBhargava:

* fixes pattern -> regex, min and max value plumbing for strings
* fixes webhook discrimination in parsing
* fixes object with only extra properties -> map type

* customer asks

* Snapshot and unit tests

### DIFF
--- a/packages/parsers/src/openapi/__test__/__snapshots__/uploadcare.json
+++ b/packages/parsers/src/openapi/__test__/__snapshots__/uploadcare.json
@@ -89,7 +89,7 @@
                 "value": {
                   "type": "primitive",
                   "value": {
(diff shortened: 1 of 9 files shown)


packages/parsers/src/openapi/__test__/__snapshots__/uploadcare.json
```diff
@@ -9597,241 +9606,6 @@
         }
       ]
     },
-    "endpoint_webhookCallbacks.fileUploaded": {
```
dsinghvi, line 9600: @RohinBhargava are these deleted endpoints intentional?

PR #2023, opened 2025-01-15
feature: add server-side feature flags

abvthecity:
- add FeatureFlags config block to navigation nodes, and evaluates it on the frontend
- enables configuring `sdk-key` for server-side evaluations
- uses the same logical and/or construct as with permissions
- continue to pass the entire navigation hierarchy to enable client-side evaluation. we won't support server-side evaluation here, because server-side pruning is an auth feature.
- use the feature flag to evaluate on the server-side whether or not to render a 404
- pass the initial flags to the frontend to bootstrap SSR'd feature flag evaluation (prevent FOUC)


dsinghvi (approved):

packages/fern-docs/bundle/src/server/ld-adapter.ts
```diff
@@ -0,0 +1,134 @@
...
+    node.featureFlags == null ||
+    node.featureFlags.length === 0 ||
+    node.featureFlags.some((flag) => {
+      const default_ = flag.default ?? false;
+      const match = flag.match ?? true;
+      return isEqual(default_, match);
+    });
+}
+
+export const createLdPredicate = async ({
```
RohinBhargava, line 105: I take it this means that the logic is that node ff are or'd. Makes sense to me!

RohinBhargava (approved):
Looks great to me! I like that the adapter makes it so that this can be extended to other providers as well, if we need to support them

PR #2028, opened 2025-01-15
fix: reduce the amount of data sent between ld server-side and ld client-side.

abvthecity:
- use a singleton to hopefully share _some_ resources between serverless function
- consume `x-secure-mode-hash` to validate the security of the server-generated flag
- disable diagnostics and events on the server-side
- disable non client-side flags (reduce payload, prevent leaking server-only flags(

### DIFF
--- a/packages/fern-docs/bundle/src/server/ld-adapter.ts
+++ b/packages/fern-docs/bundle/src/server/ld-adapter.ts
@@ -47,12 +47,13 @@ interface LaunchDarklyInfo {
   clientSideId: string;
   contextEndpoint: string;
   context: ld.LDContext | undefined;
-  defaultFlags: Record<string, unknown> | undefined;
+  defaultFlags: object | undefined;
   options:
     | {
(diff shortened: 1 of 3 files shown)


packages/fern-docs/bundle/src/server/ld-adapter.ts
```diff
@@ -133,25 +134,51 @@ export const createLdPredicate = async ({
...
   try {
-    const ldClient = ld.init(sdkKey, {
-      baseUri: options?.baseUrl,
-      streamUri: options?.streamUrl,
-      eventsUri: options?.eventsUrl,
-      stream: false,
+    const ldClient =
+      ldClientMap.get(sdkKey) ??
+      ld.init(sdkKey, {
+        baseUri: options?.baseUrl,
```
dsinghvi, line 158: not sure if this will affect perf but should we be smartly picking region?

dsinghvi (approved):

PR #2029, opened 2025-01-16
fix: ld server-side wait for init

abvthecity:


packages/fern-docs/bundle/src/server/ld-adapter.ts
```diff
@@ -152,17 +148,17 @@ async function fetchInitialFlags(
...
+    const ldClient = ld.init(sdkKey, {
+      baseUri: options?.baseUrl,
+      streamUri: options?.streamUrl,
+      eventsUri: options?.eventsUrl,
+      stream: false,
+      sendEvents: false,
+      diagnosticOptOut: true,
+    });
+
+    await ldClient.waitForInitialization({ timeout: 10 });
```
abvthecity, line 160: error logs show that this was needed

dsinghvi (approved):

PR #2032, opened 2025-01-16
fix: Correct pathing in OpenApi Parser Printing

RohinBhargava:

* There were cases where paths were bringing in extraneous `/`, which is treated as `~1` in json paths
* Add support for replacing `~` with `~0` in paths
* There were indexes like `item[$index]` -> the correct way to do this is `item/$index`. We add the notion of passing an array of paths to model this without `/` getting replaced

* Better error messaging for OAI parsing

* Logging in tests


dsinghvi (changes requested):
@RohinBhargava can you add some sort of testing on the fern check errors and make the snapshot testing capture the change in `pathId`

packages/parsers/src/openapi/__test__/__snapshots__/pathing/warnings/petstore.txt
```diff
@@ -0,0 +1,42 @@
...
+      "paths",
+      "/pet",
+      "put",
+      "security",
+    ],
+  },
+  {
+    "message": "No auth scheme found for petstore_auth. Inline security schemes are not supported.",
+    "path": [
+      "petstore",
```
dsinghvi, line 25: nit: im not sure you need the filename in the path (will make it easier to eventually render clickable links)

dsinghvi (approved):

PR #2034, opened 2025-01-16
feat: support enum with null to make nullable

RohinBhargava:

* Support nullability with enum with a nullable
* abstract optional and nullable wrapping into utility functions

* Customer rollout

* unit tests
* snapshot tests

### DIFF
--- a/packages/parsers/src/openapi/__test__/test.ts
+++ b/packages/parsers/src/openapi/__test__/test.ts
@@ -1,0 +1,50 @@
+import * as fs from "fs";
+import yaml from "js-yaml";
+import { OpenAPIV3_1 } from "openapi-types";
+import { ErrorCollector } from "../../ErrorCollector";
+import { OpenApiDocumentConverterNode } from "../3.1/OpenApiDocumentConverter.node";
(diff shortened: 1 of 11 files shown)


packages/parsers/src/openapi/__test__/test.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+// Create snapshot
+if (errors.length > 0) {
+  // console.error("errors:", errors);
+}
+// expect(errors).toHaveLength(0);
+if (warnings.length > 0) {
+  // console.warn("warnings:", warnings);
+}
```
dsinghvi, line 50: @RohinBhargava any reason you are using this as opposed to just having a snapshot here? i think its hard to understand what is fully changing but if we had checked in the monite spec here that would work just fine

RohinBhargava: This was put to bootstrap development -- the snapshot is great, will do that! I hesitate to add more customer specs for privacy reasons (actually debugged using their spec), but will begin moving into a small, targeted snapshot based approach

dsinghvi (approved):

PR #2035, opened 2025-01-16
fix(docs): preview bundle shows curl examples

dsinghvi:
For `apiV2` it would be great if the local preview showed curl examples. 

`fern docs dev` has a poor UX without these examples. 

Planning to test on dev.


dsinghvi:
Follow up ticket: https://linear.app/buildwithfern/issue/FER-4045/local-preview-shows-curl-snippets-and-more

abvthecity (approved):

PR #2039, opened 2025-01-17
chore: limit enum splaying

RohinBhargava:

* fixes an issue where very long enums were overflowing in allowed container
* The logic is if the sum of the lengths of all of the enum variants is less than 20 characters, we render the "or"s. If not, we fall back to "enum"

* UI improvement

Screenshot 2025-01-17 at 3 53 09 PM
Screenshot 2025-01-17 at 3 55 39 PM


abvthecity (approved):

packages/fern-docs/ui/src/type-shorthand/index.tsx
```diff
@@ -210,7 +211,13 @@ export function renderTypeShorthand(
       plural ? "objects" : maybeWithArticle("an", "object"),
     enum: (enumValue) => {
       // if there are only 1 or 2 values, we can list them like literals (e.g. "apple" or "banana")
-      if (enumValue.values.length > 0 && enumValue.values.length < 3) {
+      if (enumValue.values.length === 1) {
+        return `"${enumValue.values[0]?.value}"`;
+      }
+      if (
+        enumValue.values.length > 1 &&
```
abvthecity, line 218: i think we might want to just delete this logic altogether, wdyt?

PR #2040, opened 2025-01-18
fix(openrpc): respect tags and name in openrpc

dsinghvi:
This PR updates the OpenRPC converter to respect the method name and any tag assigned to the method. 

- Customer requests asking to respect name
- Customer requests asking about nesting methods in resources
- Customer requests around showing authentication

Integration tests

**Note** also removing a log line from the node collector that is affecting print statements in the CLI @abvthecity

### DIFF
--- a/packages/fdr-sdk/src/navigation/NodeCollector.ts
+++ b/packages/fdr-sdk/src/navigation/NodeCollector.ts
@@ -134,9 +134,6 @@ export class NodeCollector {
       this.orphanedNodes.push(existing.node);
       this.#setNode(node.slug, node, parents);
     } else {
-      if (FernNavigation.isPage(existing.node)) {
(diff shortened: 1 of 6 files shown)


packages/fdr-sdk/src/navigation/NodeCollector.ts
```diff
@@ -134,9 +134,6 @@ export class NodeCollector {
       this.orphanedNodes.push(existing.node);
       this.#setNode(node.slug, node, parents);
     } else {
-      if (FernNavigation.isPage(existing.node)) {
-        console.warn(`Duplicate slug found: ${node.slug}`, node.title);
-      }
```
dsinghvi, line 139: @abvthecity @RohinBhargava fyi this is creating log lines in the CLI so i am removing

PR #2024, opened 2025-01-15
feat: support local embeds

chdeskur:
This PR enables <embed> to reference locally-hosted files

Before:
Screenshot 2025-01-15 at 12 59 02 PM

After:
Screenshot 2025-01-15 at 1 00 02 PM

PR #2044, opened 2025-01-21
refactor: hide filesv2 from static props

abvthecity:
in order to support S3'ifying all markdown files, we need to remove `filesV2` from being streamed to the frontend.

before this pr, the files record is passed to the client-side and injected at runtime. after this pr, the files record is used to resolve file references on the server side.

this PR affects:
- logo
- images in markdown
- other embeds in markdown (this PR will delete `Embed` https://github.com/fern-api/fern-platform/pull/2024)


packages/fern-docs/ui/src/atoms/logo.ts
```diff
@@ -1,90 +1,107 @@
...
+//     typeof markdownText === "object"
+//       ? markdownText.frontmatter
+//       : EMPTY_FRONTMATTER;
+
+//   if (logo != null && typeof logo === "object") {
+//     if (
+//       "light" in logo &&
+//       "dark" in logo &&
+//       isFileIdOrUrl(logo.light) &&
+//       isFileIdOrUrl(logo.dark)
```
abvthecity, line 82: still working on a new home for this logic

dsinghvi (approved):
👏

PR #2048, opened 2025-01-22
Add S3 as DBDocsDefinition cache for revalidate-all/loadWithUrl/finishRegister paths

dubwub:
This will be tested on dev. Should be non-intrusive, because S3 is added as an additional cache, with Redis and RDS (existing paths) as fallbacks.


servers/fdr-deploy/scripts/fdr-deploy-stack.ts
```diff
@@ -207,6 +207,42 @@ export class FdrDeployStack extends Stack {
...
+        blockPublicAcls: false,
+        blockPublicPolicy: false,
+        ignorePublicAcls: false,
+        restrictPublicBuckets: false,
+      },
+      versioned: true,
+    });
+    dbDocsDefinitionBucket.grantPublicAccess();
+
+    const dbDocsDefinitionDomainName = `db-docs-def.${environmentType.toLowerCase()}.buildwithfern.com`;
```
dsinghvi, line 230: @dubwub i dont think our cert will allow these to work 

Can we do something like: 
https://github.com/fern-api/fern-platform/blob/5ba607a1978065765e623b6d5178e76a9937fde5/servers/fdr-deploy/scripts/fdr-deploy-stack.ts#L347-L350

`docs-definitions.buildwithfern.com` and `docs-definitions-dev2.buildwithfern.com`

servers/fdr/src/controllers/docs/v2/getDocsWriteV2Service.ts
```diff
@@ -277,6 +277,18 @@ export function getDocsWriteV2Service(app: FdrApplication): DocsV2WriteService {
           indexSegments,
         });
 
+        try {
+          await app.services.s3.writeDBDocsDefinition({
+            domain: docsRegistrationInfo.fernUrl.getFullUrl(),
+            dbDocsDefinition,
```
dsinghvi, line 283: we need to write the `readDocsDefinition`

dsinghvi (approved):

PR #2053, opened 2025-01-22
Fixing docs definition retrieval logic to use proper environment variables

dubwub:


dsinghvi (commented):
https://github.com/fern-api/fern-platform/blob/d4821148792f2dfb7cd0af7670ba5d576cb9a1af/packages/fern-docs/ui/src/services/registry.ts#L1-L20

I think we'll want to structure more like this (would look at references to `NEXT_PUBLIC_FDR_ORIGIN`)

abvthecity (approved):

PR #2107, opened 2025-02-03
fix: OpenApi Parser v2 relax no schemas constraint + reference circularity for better examples

RohinBhargava:

* Previously, the parser would return no API if schemas were not defined in the input spec
* Now, we parse even if there are no schemas being returned
* Previously, we were not checking circular references correctly
* Now we check only if a reference is circular and short-circuit at the right location, fixing pathing as a corollary

* incorrect outputs for fixtures when not including schemas

* Snapshot tests


abvthecity (approved):

packages/parsers/src/openapi/3.1/OpenApiDocumentConverter.node.ts
```diff
@@ -150,10 +150,6 @@ export class OpenApiDocumentConverterNode extends BaseOpenApiV3_1ConverterNode<
 
     const types = this.components?.convert();
 
-    if (types == null) {
```
RohinBhargava, line 153: critical change

PR #2112, opened 2025-02-04
fix: OpenApi Parser v2 thrashing nullable check

RohinBhargava:

* fixes an issue where nullable was being rewritten by checking if `nullable` is set on the object
* now, we only set nullability if `nullable: false` if already set

* bugfix

* unit tests

### DIFF
--- a/packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
+++ b/packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
@@ -79,10 +79,13 @@ export class SchemaConverterNode extends BaseOpenApiV3_1ConverterNodeWithTrackin
       accessPath: this.accessPath,
       pathId: "x-fern-availability",
     });
-    this.nullable =
-      isNonArraySchema(this.input) && isNullableSchema(this.input)
-        ? this.input.nullable
-        : undefined;
+    // only check if the schema is nullable if not already set, via nullable: true
+    if (!this.nullable) {
+      this.nullable =
+        isNonArraySchema(this.input) && isNullableSchema(this.input)
+          ? this.input.nullable
+          : undefined;
+    }
 
     // Check if the input is a reference object
     if (isReferenceObject(this.input)) {
--- a/packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
+++ b/packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
@@ -83,9 +83,7 @@ export class SchemaConverterNode extends BaseOpenApiV3_1ConverterNodeWithTrackin
     // check if nullable is set. If nullable is false, we will set it, otherwise we will ignore it
     if (isNonArraySchema(this.input) && isNullableSchema(this.input)) {
       this.nullable =
-        this.input.nullable != null && !this.input.nullable
-          ? false
-          : (this.nullable ?? this.input.nullable);
+        this.input.nullable != null ? this.input.nullable : this.nullable;
     }
 
     // Check if the input is a reference object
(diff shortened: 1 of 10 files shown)


packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -79,10 +79,13 @@ export class SchemaConverterNode extends BaseOpenApiV3_1ConverterNodeWithTrackin
...
       pathId: "x-fern-availability",
     });
-    this.nullable =
-      isNonArraySchema(this.input) && isNullableSchema(this.input)
-        ? this.input.nullable
-        : undefined;
+    // check if nullable is set. If nullable is false, we will set it, otherwise we will ignore it
+    if (isNonArraySchema(this.input) && isNullableSchema(this.input)) {
+      this.nullable =
+        this.input.nullable != null && !this.input.nullable
```
dsinghvi, line 85: Is `!this.input.nullable` a redundant check?

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
```diff
@@ -79,10 +79,13 @@ export class SchemaConverterNode extends BaseOpenApiV3_1ConverterNodeWithTrackin
...
+          ? false
+          : (this.nullable ?? this.input.nullable);
```
dsinghvi, line 87: would `this.nullable` already be set? if so, where is that happening?

packages/parsers/src/openapi/3.1/schemas/__test__/SchemaConverter.node.test.ts
```diff
@@ -222,5 +222,40 @@ describe("SchemaConverterNode", () => {
...
+        {
+          input,
+          context: mockContext,
+          accessPath: [],
+          pathId: "test",
+          seenSchemas: new Set(),
+        },
+        true
+      );
+      expect(node.nullable).toBe(true);
```
dsinghvi, line 258: why is this set to true? the input is just a `type: string` ?

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
RohinBhargava: this.input.nullable can be undefined. We do not want to set in this case -- linter filtered out the `this.input.nullable === false` explicit check, but I believe this is invariant

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
RohinBhargava: yeah, so this is when there is `type: [x, 'null']`, we strip away the null and pass nullable as a parameter to the node, recursively

packages/parsers/src/openapi/3.1/schemas/__test__/SchemaConverter.node.test.ts
RohinBhargava: `nullable: true` is passed in via the constructor

packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts
dsinghvi: `this.input.nullable != null && this.input.nullable !== undefined` would read a bit clearer

packages/parsers/src/openapi/3.1/schemas/__test__/SchemaConverter.node.test.ts
dsinghvi: https://github.com/fern-api/fern-platform/blob/182016c8e0a4dbc7c31ad63460003cc587c27bf9/packages/parsers/src/openapi/3.1/schemas/SchemaConverter.node.ts#L68 would strongly prefer that we use named arguments instead of parameters for this reason

dsinghvi (approved):

PR #2122, opened 2025-02-05
fix: inject scope into mdx-bundler

abvthecity:


RohinBhargava (approved):
Is there any quick way to test this?

PR #2126, opened 2025-02-05
fix: add language fallback for examples

chdeskur:
This PR adds a language fallback for example selection. If no example matches the currently selected language, always attempt to find a `curl` example before returning no examples. 

Fixes a bug reported by the Murf team.

### DIFF
--- a/packages/fern-docs/ui/src/api-reference/endpoints/useExampleSelection.ts
+++ b/packages/fern-docs/ui/src/api-reference/endpoints/useExampleSelection.ts
@@ -49,8 +49,11 @@ export function useExampleSelection(
           responseIndex: undefined,
         };
       }
+      console.log("language: ", language);
       const allExamples = Object.values(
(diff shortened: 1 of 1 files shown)


packages/fern-docs/ui/src/api-reference/endpoints/useExampleSelection.ts
```diff
@@ -49,8 +49,11 @@ export function useExampleSelection(
           responseIndex: undefined,
         };
       }
+      console.log("language: ", language);
```
fern-support, line 52: delete console.log

fern-support (approved):

packages/fern-docs/ui/src/api-reference/endpoints/useExampleSelection.ts
chdeskur: valid

PR #2128, opened 2025-02-05
fix: Display nulls properly and clean up example generation logic

RohinBhargava:

* Previously, fallback examples were added to everything. Now we check to see that there are no other examples before falling back
* Previously if a primitive was nullable, we were returning `"null"`. We actually want to return `null`

* Customer asks

* Snapshot tests


packages/parsers/src/openapi/3.1/paths/response/ResponseMediaTypeObjectConverter.node.ts
```diff
@@ -178,45 +360,15 @@ export class ResponseMediaTypeObjectConverterNode extends BaseOpenApiV3_1Convert
...
-                    )
-                  : undefined
-              )
-            )
-            .filter(isNonNullish)
-        )
-    );
+    this.examples ??= [];
+
+    const foundExampleNames = this.matchExamplesByName(responseExamples);
```
dsinghvi, line 365: `const { matchingExampleNames } = this.computeMatchingRequestResponseExamples();`

packages/parsers/src/openapi/3.1/paths/response/ResponseMediaTypeObjectConverter.node.ts
```diff
@@ -178,45 +360,15 @@ export class ResponseMediaTypeObjectConverterNode extends BaseOpenApiV3_1Convert
...
+    this.matchExamplesByIndex(foundExampleNames, responseExamples);
```
dsinghvi, line 366: `const examplePairs = this.matchExamplesByIndex(uncorrelatedExamples)`

dsinghvi (approved):

PR #2129, opened 2025-02-05
fix: OneOf and AnyOf now support more than just object variants

RohinBhargava:

* Previously, oneOfs and anyOfs were constrained to only having object variants. This is wrong

* Customer fixes

* Snapshot test


packages/parsers/src/openapi/3.1/schemas/OneOfConverter.node.ts
```diff
@@ -34,10 +34,16 @@ export class OneOfConverterNode extends BaseOpenApiV3_1ConverterNodeWithTracking
...
-          resolveSchemaReference(schema, this.context.document)?.type ===
-          "object"
+      this.isUnionOfContainers = (this.input.oneOf ?? this.input.anyOf)?.every(
+        (schema) => {
+          const resolvedSchema = resolveSchemaReference(
+            schema,
+            this.context.document
+          );
+          return (
+            resolvedSchema?.type === "object" || resolvedSchema?.enum != null
```
dsinghvi, line 44: why are you checking this?

dsinghvi (approved):

PR #2130, opened 2025-02-05
feat: support exploded form parameters in docs

dsinghvi:
Updates form data parameters to have an option for being `exploded`. It also allows the example to be exploded as well. 

- Customer demand
- Curl snippets in the API Playground should be accurate
- Playground calls should work as well

- Unit tests in the PR


RohinBhargava (approved):

packages/fern-docs/ui/src/playground/utils/serialize.ts
```diff
@@ -53,6 +53,25 @@ export async function serializeFormStateBody(
...
+            if (property?.exploded) {
+              // For exploded form fields, convert value to array if not already
+              const arrayValue = Array.isArray(value.value)
+                ? value.value
+                : [value.value];
+              formDataValue[key] = {
+                type: "json",
+                value: arrayValue,
+                contentType,
+              };
```
dsinghvi, line 71: this is a bug -- we actually need to update and introduce an `exploded` parameter here

RohinBhargava (approved):
Looks good to me! Not sure if I follow the bug, but looks great otherwise

packages/fern-docs/ui/src/playground/fetch-utils/requestToBodyInit.ts
```diff
@@ -51,6 +51,26 @@ export async function toBodyInit(
               }
             }
             break;
+          case "exploded":
```
RohinBhargava, line 54: Sweet!

PR #2154, opened 2025-02-10
feat: override matching for code blocks

chdeskur:
This PR introduces the ability to override the matching language for CodeBlocks:

https://github.com/user-attachments/assets/0249529a-c270-4d33-9d98-2a4930ef2b06

Usage is as follows:
````md
<CodeGroup>
```bash NPM matchLanguage="typescript"
npm install fern-api
```
```bash PyPI matchLanguage="python"
pypi install fern
```
</CodeGroup>
````

### DIFF
--- a/packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx
+++ b/packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx
@@ -21,6 +21,7 @@ export interface FernSyntaxHighlighterProps {
   viewportRef?: React.RefObject<ScrollToHandle>;
   maxLines?: number;
   wordWrap?: boolean;
+  matchLanguage?: string;
 }
 
 export const FernSyntaxHighlighter = forwardRef<
@@ -38,7 +39,7 @@ export const FernSyntaxHighlighter = forwardRef<
       return highlightTokens(highlighter, code, language);
     } catch (e) {
       // TODO: sentry
-      // eslint-disable-next-line no-console
+
       console.error("Error occurred while highlighting tokens", e);
       return createRawTokens(code, language);
     }
--- a/packages/fern-docs/ui/src/mdx/components/code/CodeGroup.tsx
+++ b/packages/fern-docs/ui/src/mdx/components/code/CodeGroup.tsx
@@ -35,15 +35,8 @@ export const CodeGroup: React.FC<React.PropsWithChildren<CodeGroup.Props>> = ({
   useEffect(() => {
     if (selectedLanguage) {
       const matchingTab = itemsRef.current.find((item) => {
-        if (item.matchLanguage) {
-          return (
-            ApiDefinition.cleanLanguage(item.matchLanguage) === selectedLanguage
-          );
-        }
-        return (
-          item.language &&
-          ApiDefinition.cleanLanguage(item.language) === selectedLanguage
-        );
+        const matchLanguage = item.matchLanguage || item.language;
+        return ApiDefinition.cleanLanguage(matchLanguage) === selectedLanguage;
       });
 
       if (matchingTab) {
@@ -51,12 +44,10 @@ export const CodeGroup: React.FC<React.PropsWithChildren<CodeGroup.Props>> = ({
         setSelectedTabIndex((prevIndex) => {
           const prevTab = itemsRef.current[prevIndex];
           if (prevTab) {
-            const currentMatch = prevTab.matchLanguage
-              ? ApiDefinition.cleanLanguage(prevTab.matchLanguage) ===
-                selectedLanguage
-              : prevTab.language &&
-                ApiDefinition.cleanLanguage(prevTab.language) ===
-                  selectedLanguage;
+            const prevMatchLanguage = prevTab.matchLanguage || prevTab.language;
+            const currentMatch =
+              ApiDefinition.cleanLanguage(prevMatchLanguage) ===
+              selectedLanguage;
             if (currentMatch) {
               return prevIndex;
             }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx [hunk 1]
```diff
@@ -21,6 +21,7 @@ export interface FernSyntaxHighlighterProps {
   viewportRef?: React.RefObject<ScrollToHandle>;
   maxLines?: number;
   wordWrap?: boolean;
+  matchLanguage?: string;
```
abvthecity, line 24: Why at matchLanguage? Let's just use the language prop above

packages/fern-docs/ui/src/mdx/plugins/rehypeFernCode.ts
```diff
@@ -279,6 +284,12 @@ export function parseBlockMetaString(
     meta = meta.replace(wordWrap[0], "");
   }
 
+  const matchLanguage = meta.match(
+    /matchLanguage=(?:"((?:[^"\\]|\\.)*?)"|'((?:[^'\\]|\\.)*?)')/
```
abvthecity, line 288: Don't think we should add this

abvthecity (changes requested):

packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx
chdeskur: i think language needs to be used for syntax highlighting?

packages/fern-docs/ui/src/mdx/plugins/rehypeFernCode.ts
chdeskur: would you just match (/s|/S)*?

packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx
abvthecity: The language for syntax highlighting is the same language we want to use to save the user's language preference

abvthecity: But only when the user explicitly navigates to it

abvthecity: Actually scratch that. I think I understand why this is necessary (the bash example)

Can you fallback matchedLanguage to language if not specified?

chdeskur: was handling the fallback in CodeGroup.tsx, does it make more sense to do it in rehypeFernCode?

packages/fern-docs/ui/src/mdx/plugins/rehypeFernCode.ts
chdeskur: i just copied the parsing from `title` to make sure we support both ' and "

packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx
abvthecity: (I'll point it out below)

packages/fern-docs/ui/src/mdx/components/code/CodeGroup.tsx
```diff
@@ -34,19 +35,31 @@ export const CodeGroup: React.FC<React.PropsWithChildren<CodeGroup.Props>> = ({
   useEffect(() => {
     if (selectedLanguage) {
       const matchingTab = itemsRef.current.find((item) => {
-        const normalizedLanguage = ApiDefinition.cleanLanguage(item.language);
-        return item.language && normalizedLanguage === selectedLanguage;
+        if (item.matchLanguage) {
```
abvthecity, line 38: Over here, matchLanguage = item.matchLanguage || item.language is more explicit and clean

packages/fern-docs/ui/src/mdx/components/code/CodeGroup.tsx
```diff
@@ -59,9 +72,11 @@ export const CodeGroup: React.FC<React.PropsWithChildren<CodeGroup.Props>> = ({
     setSelectedTabIndex(newIndex);
 
     const tab = itemsRef.current[newIndex];
-    const normalizedLanguage = tab?.language
-      ? ApiDefinition.cleanLanguage(tab.language)
-      : undefined;
+    const normalizedLanguage = tab?.matchLanguage
+      ? ApiDefinition.cleanLanguage(tab.matchLanguage)
+      : tab?.language
```
abvthecity, line 77: Here too

packages/fern-docs/syntax-highlighter/src/FernSyntaxHighlighter.tsx
chdeskur: ahh i think i know what you mean - it would be cleaner to always assume we're opening the code block based on a single prop, so it could be better to use matchLanguage as that prop, rather than check if matchLanguage exists otherwise use language

packages/fern-docs/ui/src/mdx/plugins/rehypeFernCode.ts
```diff
@@ -306,6 +317,7 @@ export function parseBlockMetaString(
     focused: focused != null,
     wordWrap: wordWrap != null,
     highlights,
+    matchLanguage: match,
```
abvthecity, line 320: I think "for" is a better syntactical sugar

Like 

````

```bash for=python

````

chdeskur: addressed in parsing - i like matchLanguage for variable use since it makes it clear this has to do with the language property

abvthecity (approved):

PR #2161, opened 2025-02-12
[DO NOT MERGE]: accordion scrolling update

chdeskur:
This PR proposes a solution for accordion scrolling (for linking directly within an accordion, scrolling when TOC is selected)


chdeskur:
change the approach to prefix-based matching

PR #2164, opened 2025-02-13
fix: Fix endpoint and subpackage ids for better ergonomics with Nav Converter in CLI and objects

RohinBhargava:

* Changes the output ids for endpoints, webhooks and subpackages to better conform with Nav converter in CLI
* Respect default optional object parameters
* Begin to transition to kwargs for extra arguments
* Add names for undiscriminated union variants

* Go lives

Snapshots

### DIFF
--- a/packages/parsers/src/openapi/3.1/paths/OperationObjectConverter.node.ts
+++ b/packages/parsers/src/openapi/3.1/paths/OperationObjectConverter.node.ts
@@ -10,6 +10,7 @@ import { coalesceServers } from "../../utils/3.1/coalesceServers";
 import { resolveParameterReference } from "../../utils/3.1/resolveParameterReference";
 import { getEndpointId } from "../../utils/getEndpointId";
 import { getExampleName } from "../../utils/getExampleName";
+import { maybeSingleValueToArray } from "../../utils/maybeSingleValueToArray";
 import { mergeSnippets } from "../../utils/mergeSnippets";
 import { mergeXFernAndResponseExamples } from "../../utils/mergeXFernAndResponsesExamples";
 import { SecurityRequirementObjectConverterNode } from "../auth/SecurityRequirementObjectConverter.node";
@@ -33,10 +34,10 @@ import { ResponsesObjectConverterNode } from "./response/ResponsesObjectConverte
 
 export class OperationObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
   OpenAPIV3_1.OperationObject,
-  | FernRegistry.api.latest.EndpointDefinition
-  | FernRegistry.api.latest.WebhookDefinition
+  | FernRegistry.api.latest.EndpointDefinition[]
+  | FernRegistry.api.latest.WebhookDefinition[]
 > {
-  endpointId: string | undefined;
+  endpointIds: string[] | undefined;
   description: string | undefined;
   displayName: string | undefined;
   operationId: string | undefined;
@@ -47,7 +48,7 @@ export class OperationObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
   responses: ResponsesObjectConverterNode | undefined;
   availability: AvailabilityConverterNode | undefined;
   auth: SecurityRequirementObjectConverterNode | undefined;
-  namespace: XFernGroupNameConverterNode | undefined;
+  namespaces: XFernGroupNameConverterNode[] | undefined;
   xFernExamplesNode: XFernEndpointExampleConverterNode | undefined;
   redocExamplesNode: RedocExampleConverterNode | undefined;
   emptyResponseExamples: ExampleObjectConverterNode[] | undefined;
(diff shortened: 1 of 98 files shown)


packages/parsers/src/openapi/utils/getEndpointId.ts
```diff
@@ -20,5 +21,11 @@ export function getEndpointId(
...
     return undefined;
   }
-  return `${isWebhook ? "webhook_" : "endpoint_"}${camelCase(namespace != null ? (typeof namespace === "string" ? namespace : namespace.join("_")) : "")}.${camelCase(sdkMethodName ?? "") || operationId || camelCase(endpointName)}`;
+  return `${isWebhook ? "subpackage_" : "endpoint_"}${
+    namespace != null
+      ? maybeSingleValueToArray(namespace)
+          ?.map((member) => camelCase(member))
+          .join(".")
+      : ""
+  }.${camelCase(sdkMethodName ?? "") || operationId || camelCase(endpointName)}`;
```
dsinghvi, line 30: can we break these into helper functions that are little bit easier to understand? 

```ts
const prefix = getPrefix();
const namespace = getNamespaceIdComponent();
const endpoint = getEndpointNameComponent(); 
return `${prefix}.${namespace}.${endpoint}`


function getPrefix():  string {

}

function getNamespaceIdComponent(): string {

}

function getEndpointNameComponent(): string {

}
```

abvthecity (approved):

packages/parsers/src/openapi/3.1/paths/OperationObjectConverter.node.ts
```diff
@@ -315,14 +332,18 @@ export class OperationObjectConverterNode extends BaseOpenApiV3_1ConverterNode<
...
-      this.namespace?.groupName,
-      this.path,
-      this.method,
-      sdkMethodName.sdkMethodName,
-      this.input.operationId,
-      this.isWebhook
-    );
+    this.endpointIds = this.namespaces
+      .map((namespace) =>
+        getEndpointId(
```
dsinghvi, line 337: can you update this function to take in an object instead of parameters

servers/fdr/src/controllers/api/getRegisterApiService.ts
```diff
@@ -110,16 +111,13 @@ export function getRegisterApiService(app: FdrApplication): APIV1WriteService {
...
-          (req.body.definitionV2 != null &&
-            Object.keys(req.body.definitionV2).length === 0)
-        ) {
-          throw new Error("No latest definition provided");
-        }
-        transformedApiDefinition = req.body.definitionV2;
+      if (req.body.definitionV2 != null && transformedApiDefinition != null) {
+        transformedApiDefinition = mergeEndpointSnippets(
+          req.body.definitionV2,
+          transformedApiDefinition as APIV1Db.DbApiDefinition
```
dsinghvi, line 119: why do we have to do a cast here?

servers/fdr/src/controllers/api/getRegisterApiService.ts
```diff
@@ -110,16 +111,13 @@ export function getRegisterApiService(app: FdrApplication): APIV1WriteService {
...
-      if (transformedApiDefinition == null) {
-        if (
-          req.body.definitionV2 == null ||
-          (req.body.definitionV2 != null &&
-            Object.keys(req.body.definitionV2).length === 0)
-        ) {
-          throw new Error("No latest definition provided");
-        }
-        transformedApiDefinition = req.body.definitionV2;
+      if (req.body.definitionV2 != null && transformedApiDefinition != null) {
```
dsinghvi, line 116: does this mean the CLI has to ship up both `definition` and `definitionV2` ?

servers/fdr/src/controllers/api/getRegisterApiService.ts
RohinBhargava: This can be one of two types, and it feels wrong to create a guard for a temporary solution

servers/fdr/src/controllers/api/getRegisterApiService.ts
RohinBhargava: yes, unfortunately

packages/parsers/src/openapi/utils/getEndpointId.ts
```diff
@@ -20,5 +21,11 @@ export function getEndpointId(
   if (endpointName == null) {
     return undefined;
   }
-  return `${isWebhook ? "webhook_" : "endpoint_"}${camelCase(namespace != null ? (typeof namespace === "string" ? namespace : namespace.join("_")) : "")}.${camelCase(sdkMethodName ?? "") || operationId || camelCase(endpointName)}`;
+  return `${isWebhook ? "subpackage_" : "endpoint_"}${
```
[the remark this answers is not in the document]
RohinBhargava: yeah, I was surprised by this, but v1 calls them `subpackage_...`

servers/fdr/src/controllers/api/getRegisterApiService.ts
RohinBhargava: will split into two variables instead!

PR #2167, opened 2025-02-14
chore: Add snippetsConfiguration to ApiDefinitionLatest

RohinBhargava:

* Adds snippetsConfiguration to `ApiDefinitionLatest` payload, so that endpoints can be enriched with snippets
* Adds enrichment logic in registration step. Uses the same code paths as legacy `ApiDefinition`, so there should be little to no drift

* Sdk snippets for OpenApi v2 Parser customers

* To be tested with dev cli

### DIFF
--- a/fern/apis/fdr/definition/api/latest/__package__.yml
+++ b/fern/apis/fdr/definition/api/latest/__package__.yml
@@ -8,7 +8,7 @@ imports:
   commons: commons.yml
   rootCommons: ../../commons.yml
   auth: auth.yml
-
+  v1Register: ../v1/register/__package__.yml
 service:
   base-path: /registry/api/latest
   auth: true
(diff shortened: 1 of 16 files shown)


fern/apis/fdr/definition/api/latest/__package__.yml
```diff
@@ -35,6 +35,7 @@ types:
       subpackages: map<commons.SubpackageId, commons.SubpackageMetadata>
       auths: map<auth.AuthSchemeId, auth.AuthScheme>
       globalHeaders: optional<list<type.ObjectProperty>>
+      snippetsConfiguration: optional<v1Register.SnippetsConfig>
```
dsinghvi, line 38: can we create a new object called `sdks` ?

dsinghvi: And redefine the object?

dsinghvi (approved):

PR #2172, opened 2025-02-17
chore: fail gracefully on endpoint url display

chdeskur:
This PR handles some of the possibly null values in the Environment to fail more gracefully when protocols or hosts aren't set. 

Before:
Screenshot 2025-02-17 at 12 32 19 PM

After:
Screenshot 2025-02-17 at 12 30 52 PM

### DIFF
--- a/packages/fern-docs/ui/src/components/MaybeEnvironmentDropdown.tsx
+++ b/packages/fern-docs/ui/src/components/MaybeEnvironmentDropdown.tsx
@@ -81,12 +81,18 @@ export function MaybeEnvironmentDropdown({
     parse(inputValue).host != null &&
     parse(inputValue).protocol != null;
 
-  const urlProtocol = url ? url.protocol : "";
-  const fullyQualifiedDomainAndBasePath = url
-    ? url.pathname != null && url.pathname !== "/"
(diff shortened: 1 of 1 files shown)


packages/fern-docs/ui/src/components/MaybeEnvironmentDropdown.tsx
```diff
@@ -81,12 +81,18 @@ export function MaybeEnvironmentDropdown({
...
+  const fullyQualifiedDomainAndBasePath =
+    typeof url === "string"
+      ? null
+      : url
+        ? url.pathname != null && url.pathname !== "/"
+          ? url.host
+            ? `${url.host}${url.pathname}`
+            : url.pathname
+          : url.host
+        : null;
```
dsinghvi, line 95: nit: i would probably use basic `if` statements hear than the ternary operator since its hard to grok at first glance what is happening:

dsinghvi (approved):

packages/fern-docs/ui/src/components/MaybeEnvironmentDropdown.tsx
chdeskur: valid

PR #2176, opened 2025-02-17
fix: OpenApi Parser v2 Conform to legacy ids for snippet matching

RohinBhargava:

* Previously, we generated a new version of unique ids for endpoints, so that they were cleaner in the resulting payload. Now we generate the legacy apis
* Previously, we only tried to match on endpoint path -- upon closer inspection, it looks like the stringifying function appended an erroneous `/` and did not conform to the path parameter specification as in the table. Now we stringify properly, and use the legacy id to be more robust when matching

* Make sure that snippets properly match

*. To be tested with dev cli

### DIFF
--- a/packages/parsers/src/openapi/3.1/auth/__test__/OAuth2SecuritySchemeConverter.node.test.ts
+++ b/packages/parsers/src/openapi/3.1/auth/__test__/OAuth2SecuritySchemeConverter.node.test.ts
@@ -35,7 +35,7 @@ describe("OAuth2SecuritySchemeConverterNode", () => {
         type: "clientCredentials",
         value: {
           type: "referencedEndpoint",
-          endpointId: "endpoint_.postHttpsApiExampleComOauthToken",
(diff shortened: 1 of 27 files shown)


packages/parsers/src/openapi/3.1/auth/__test__/OAuth2SecuritySchemeConverter.node.test.ts
```diff
@@ -35,7 +35,7 @@ describe("OAuth2SecuritySchemeConverterNode", () => {
         type: "clientCredentials",
         value: {
           type: "referencedEndpoint",
-          endpointId: "endpoint_.postHttpsApiExampleComOauthToken",
+          endpointId: "endpoint_.post-https-api-example-com-oauth-token",
```
dsinghvi, line 38: these ids still look wrong

dsinghvi (approved):

PR #2177, opened 2025-02-17
[DRAFT] error handling and notification for AI search failures

prb2:
Fixes FER-4445

Added a `getErrorMessage` handler to the `streamText` result's `toDataStreamResponse`. The handler makes a best effort to identify the error kind and then logs an error message and posts to the #engineering-notifs slack channel.

Users and staff have noticed non-deterministic failures of the `Ask AI` feature where an reasonable prompt returns an empty response. This change will give us some idea of how frequently this issue is occurring.

**Local testing without Slack Token**

Ran locally with `pnpm docs:dev` and repro'd the bug. Confirmed that the response contained an error message with Log ID:

Screenshot 2025-02-17 at 4 59 46 PM

**Local testing with Slack Token**
- Slack message
Screenshot 2025-02-17 at 7 09 07 PM

- console log
Screenshot 2025-02-17 at 7 09 31 PM

### DIFF
--- a/packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
+++ b/packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
@@ -1,4 +1,5 @@
 import { track } from "@/server/analytics/posthog";
+import { WebClient } from "@slack/web-api";
 import { safeVerifyFernJWTConfig } from "@/server/auth/FernJWT";
 import { getOrgMetadataForDomain } from "@/server/auth/metadata-for-url";
 import { openaiApiKey, turbopufferApiKey } from "@/server/env-variables";
@@ -12,11 +13,12 @@ import {
   toDocuments,
 } from "@fern-docs/search-server/turbopuffer";
 import { COOKIE_FERN_TOKEN, withoutStaging } from "@fern-docs/utils";
-import { embed, EmbeddingModel, streamText, tool } from "ai";
+import { embed, EmbeddingModel, InvalidToolArgumentsError, NoSuchToolError, streamText, tool, ToolExecutionError } from "ai";
 import { initLogger, wrapAISDKModel } from "braintrust";
 import { cookies } from "next/headers";
 import { NextRequest, NextResponse } from "next/server";
 import { z } from "zod";
+import { v4 as uuid } from "uuid";
 
 export const maxDuration = 60;
 export const revalidate = 0;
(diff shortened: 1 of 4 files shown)


packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
```diff
@@ -132,7 +134,44 @@ export async function POST(req: NextRequest) {
...
+        errorKind = "ToolExecutionError"
+      }
+      const logId = uuid()
+      const basicMsg  = `An internal error occurred. Log ID: ${ logId }`;
+      console.error(basicMsg)
+
+      const slackToken = process.env.SLACK_TOKEN;
+      if (slackToken) {
+        const slackMsg = basicMsg + `${errorKind} ${error}`
+        const webClient = new WebClient(slackToken);
```
prb2, line 160: This is for illustrative purposes only...wasn't sure where I should actually create the slack client. Maybe in some `util` module somewhere in `packages/fern-docs/bundle/src/app/`? @abvthecity any suggestions?

I saw https://github.com/fern-api/fern-platform/blob/main/servers/fdr/src/services/slack/SlackService.ts but that looks like a separate service? Wasn't sure how to import and reuse that code here.

packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
```diff
@@ -132,7 +134,44 @@ export async function POST(req: NextRequest) {
...
+        webClient.chat.postMessage({
+          channel: "#engineering-notifs",
+          text: slackMsg,
+        });
+      } else {
+        // if we can't post to slack, just log the full error to the console
+        // TODO: revisit; we may not want to show the user this error in the console
+        console.error(error)
```
prb2, line 168: What do the reviewers think? Is it too risky to log the full error to the console in case we can't post to slack?

packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
```diff
@@ -132,7 +134,44 @@ export async function POST(req: NextRequest) {
...
+        errorKind = "NoSuchToolError"
+      }
+      else if (InvalidToolArgumentsError.isInstance(error)) {
+        errorKind = "InvalidToolArgumentsError"
+      }
+      else if (ToolExecutionError.isInstance(error)) {
+        errorKind = "ToolExecutionError"
+      }
+      const logId = uuid()
+      const basicMsg  = `An internal error occurred. Log ID: ${ logId }`;
```
prb2, line 154: I added this Log ID so that if a customer saw the error message in the console, they could reference it when talking with us to get support. Not sure how important this is or if there's some better way to do it...lmk

dsinghvi: That sounds great

dsinghvi: This entire functions runs as a serverless function in AWS (via Vercel) -- so nothing here will post in the console

prb2: Got it; removed the log ID.

packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
prb2: Simplified since this will only get logged in vercel.

packages/fern-docs/bundle/src/app/api/fern-docs/search/v2/chat/route.ts
prb2: Since this is all running in a serverless invocation, I don't think we can avoid creating a slack web client every time we need to emit this notification. At best, we can factor out this code into a helper function so we don't need to copy it everytime

dsinghvi: Yup this sounds good -- no need to refactor from my pov, if you want feel free to put it into a `slack.ts` that we can then move around

dsinghvi (approved):

abvthecity (approved):

PR #2181, opened 2025-02-18
docs/ask-ai: tell user to retry when error occurs

prb2:

Provide an `onError` handler to `useChat` that updates a bool atom. This state can then be used to show a useful error message rather than a blank response.

It's useful to show users a message rather than a blank response when an error occurs.

Tested locally: invoked an error and verified that the new error message was displayed.

Screenshot 2025-02-18 at 3 25 38 PM

### DIFF
--- a/packages/fern-docs/search-ui/src/components/desktop/desktop-ask-ai.tsx
+++ b/packages/fern-docs/search-ui/src/components/desktop/desktop-ask-ai.tsx
@@ -249,6 +249,7 @@ const DesktopAskAIContent = (props: {
 };
 
 const initialConversationAtom = atom<Message[]>([]);
+const chatErrorAtom = atom(false);
 
 const DesktopAskAIChat = ({
   onReturnToSearch,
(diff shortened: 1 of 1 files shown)


packages/fern-docs/search-ui/src/components/desktop/desktop-ask-ai.tsx
prb2, line 1: I'm new to typescript/react/jotai state management...let me know if there's a better way to go about this!

abvthecity (changes requested):
can you include a retry button? it's better UX to let the user click-to-retry without having to retype the entire query again.

also do we know anything about the throttling rate? maybe a countdown to retry in 10 seconds?

prb2:
> can you include a retry button? it's better UX to let the user click-to-retry without having to retype the entire query again.

Agreed, that would definitely be a better UX. 

@abvthecity What do you think about merging this PR as is to get the quick win of not having an empty response, and I can file a ticket for adding the retry button + countdown? Asking because I think it might take me some time to do this, and I'd rather continue making progress on the broken link checker tasks I have. Let me know, happy to go either way!

abvthecity (approved):

abvthecity:
Makes sense. Approved to unblock, would love to revisit this in the future, maybe with an automatic retry. Can you loop in @dubwub on increasing limits?

PR #2186, opened 2025-02-19
fix: accordion behavior

chdeskur:
This PR queues up the change in accordion behavior to allow for deep-linking within accordions.


packages/fern-docs/ui/src/mdx/bundlers/mdx-bundler.ts
```diff
@@ -147,11 +147,17 @@ async function serializeMdxImpl(
           {
             mdxJsxElementAllowlist: {
               Frame: ["caption"],
-              Tab: ["title"],
-              Card: ["title", "icon"],
-              Callout: ["title", "icon"],
-              Step: ["title"],
-              Accordion: ["title"],
+              Tab: ["title", "id"],
```
abvthecity, line 150: the intent of the mdxJsxElementAllowlist is to specifically allow using JSX inside of those props (for example, adding a <bold> tag in the title, or using an  tag in icon). ID should always be a string.

packages/fern-docs/ui/src/mdx/bundlers/next-mdx-remote.ts
```diff
@@ -69,7 +69,13 @@ function withDefaultMdxOptions(
           Card: ["title", "icon"],
           Callout: ["title", "icon"],
           Step: ["title"],
-          Accordion: ["title"],
+          Accordion: ["title", "id"],
+          h1: ["id"],
+          h2: ["id"],
+          h3: ["id"],
```
chdeskur, line 75: will remove this as well

chdeskur:
https://github.com/user-attachments/assets/a1c4000c-872e-4442-a140-50c7a539afb3

abvthecity (approved):

PR #2194, opened 2025-02-20
fix: OpenApi Parser v2 return auths if types are nullish

RohinBhargava:

* Previously, auths were not propagating if there were no schemas. Now they do

* Customer edge case

* Snapshot test


packages/parsers/src/openapi/3.1/OpenApiDocumentConverter.node.ts
```diff
@@ -138,6 +138,7 @@ export class OpenApiDocumentConverterNode extends BaseOpenApiV3_1ConverterNode<
     > = computeSubpackages({ endpoints, webhookEndpoints: webhooks });
 
     const { types, auths } = this.components?.convert() ?? {};
+    console.log(types, auths);
```
Swimburger, line 141: intentional log?

Swimburger (approved):

PR #2208, opened 2025-02-23
fix: image ratios

chdeskur:
This PR adds logic to handle images where the `height` _or_ `width` property is set, but not both. 

My favorite before:
Screenshot 2025-02-23 at 2 55 00 PM
Screenshot 2025-02-23 at 2 55 16 PM
Screenshot 2025-02-23 at 2 55 25 PM

And after:
Screenshot 2025-02-23 at 2 55 32 PM


packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -74,9 +74,15 @@ export function rehypeFiles(
           height != null &&
           !attributes.find((attr) => attr.name === "height")
         ) {
+          const actualWidth = Number(
```
fern-support, line 77: nit: in order to avoid doing the `find` operation twice, you can declare it above the `if` condition

abvthecity, line 77: nit: to avoid calling `find()` twice, you can declare it above the `if` condition.

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -74,9 +74,15 @@ export function rehypeFiles(
...
+            attributes.find((attr) => attr.name === "width")?.value ?? 0
+          );
+          const adjustedHeight =
+            actualWidth && width ? height * (actualWidth / width) : height;
+
           node.attributes.unshift({
             name: "height",
-            value: String(height),
+            value: String(Math.round(adjustedHeight)),
```
abvthecity, line 85: need to remove rounding because fractional units should be allowed

chdeskur:
take 2: 

before: https://image-demo.docs.buildwithfern.com/image-demo

after: 
Screenshot 2025-02-25 at 4 15 06 PM

Screenshot 2025-02-25 at 4 15 12 PM

Screenshot 2025-02-25 at 4 15 16 PM

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -191,3 +172,65 @@ function getEstree(
...
+    if (!attrHeight) {
+      addStyle = `width:${attrWidth.value}px;`;
+    } else {
+      addStyle = `width:${attrWidth.value}px;height:${attrHeight.value};`;
+    }
+  } else if (attrHeight) {
+    addStyle = `height:${attrHeight.value}px;width:auto;`;
+  }
+
+  node.attributes.unshift(unknownToMdxJsxAttribute("imageSize", addStyle));
```
abvthecity, line 210: instead of running `parseStringStyle` inside of the types, lets just assemble the `{}` and pass it into unknownToMdxJsxAttribute.

abvthecity (approved):

packages/fern-docs/ui/src/mdx/__test__/__snapshots__/rehypeFernFiles.test.ts.snap
```diff
@@ -0,0 +1,329 @@
...
+    {
+      "name": "imageSize",
+      "type": "mdxJsxAttribute",
+      "value": {
+        "data": {
+          "estree": {
+            "body": [
+              {
+                "expression": {
+                  "properties": [],
```
abvthecity, line 302: this is a bit odd i'd expect there to be properties here?

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -191,3 +172,71 @@ function getEstree(
   }
   return null;
 }
+
+export function setDimension(
+  node: Hast.MdxJsxElement,
+  attributes: MdxJsxAttribute[],
+  width: number | undefined,
```
abvthecity, line 179: super nit:
calling this `intrinsicWidth and intrinsicHeight` would be extra clear

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -191,3 +172,71 @@ function getEstree(
...
+    } else {
+      addStyle = {
+        width: `${attrWidth.value}px`,
+        height: `${attrHeight.value}px`,
+      };
+    }
+  } else if (attrHeight) {
+    addStyle = {
+      height: `${attrHeight.value}px`,
+      width: "auto",
```
abvthecity, line 212: doesn't the aspectRatio already take care of the width? might be missing something here.

chdeskur: was reading this: https://nextjs.org/docs/app/api-reference/components/image#style

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -191,3 +172,71 @@ function getEstree(
...
+  // validate width and height are numeric
+  if (
+    (attrWidth?.value && isNaN(Number(attrWidth.value))) ||
+    (attrHeight?.value && isNaN(Number(attrHeight.value)))
+  ) {
+    return;
+  }
+
+  let addStyle: React.CSSProperties = {};
+  if (attrWidth) {
```
abvthecity, line 200: nit: this needs a bit of refactoring for code clarity but on the right track

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
chdeskur: love it

packages/fern-docs/ui/src/mdx/__test__/__snapshots__/rehypeFernFiles.test.ts.snap
chdeskur: working through a weird rebase, will look into this

packages/fern-docs/ui/src/mdx/components/html/image.tsx
```diff
@@ -41,6 +42,7 @@ export const Image = forwardRef<
     loading?: "eager" | "lazy" | undefined;
     blurDataURL?: string | undefined;
     unoptimized?: boolean | undefined;
+    imageSize?: CSSProperties | undefined;
```
abvthecity, line 45: nit: what i like to do to tell someone _not_ to use the property is to append it with a `__` at the begining like:

`__do_not_set_imageSizeStyle` with a comment about why it exists.

see here: https://github.com/fern-api/fern-platform/blob/4558e052d91c36298dc2101a6b2b44e91413e0d2/packages/fern-docs/ui/src/atoms/auth.ts#L19

just to signify this is a non-standard utility being generated / handled

packages/fern-docs/ui/src/mdx/__test__/__snapshots__/rehypeFernFiles.test.ts.snap
chdeskur: so the imageSize is only set if the intrinsic size needs to be set using inline styling.

packages/fern-docs/ui/src/mdx/plugins/rehype-files.ts
```diff
@@ -191,3 +172,62 @@ function getEstree(
...
+  ) {
+    return;
+  }
+
+  const addStyle: React.CSSProperties = {
+    ...(attrWidth && { width: `${attrWidth.value}px` }),
+    ...(attrHeight && { height: `${attrHeight.value}px` }),
+    ...(!attrWidth && attrHeight && { width: "auto" }),
+  };
+
```
chdeskur, line 204: @abvthecity code above is refactored for clarify, code below demonstrates the renamed attribute

abvthecity: looks great!

packages/fern-docs/ui/src/mdx/__test__/__snapshots__/rehypeFernFiles.test.ts.snap
abvthecity: maybe __assigned_imageSize is being set even though there's no overrides from the user being set?

chdeskur: yep, i added a condition to avoid adding an empty __assigned_imageSize

PR #2268, opened 2025-03-10
fix: check for nan

chdeskur:
ensure iconSize is a number

before: 
Screenshot 2025-03-10 at 8 56 32 AM

after: 
Screenshot 2025-03-10 at 8 57 46 AM


abvthecity (changes requested):
The default should be set on the component itself and not via rehype plugin

PR #2334, opened 2025-03-14
fix: add fallback url when `servers` is not defined

matte-fern:
Fixes FER-4654

- Added a check for pruned endpoints
- Added default entry `https://host.com` if `environments` is empty

- Fix breaking docs when `servers` not defined in `openapi.json`

- tested locally

### DIFF
--- a/packages/fern-docs/bundle/src/server/docs-loader.ts
+++ b/packages/fern-docs/bundle/src/server/docs-loader.ts
@@ -347,10 +347,10 @@ const createGetPrunedApiCached = (domain: string) =>
 
       for (const endpointK of Object.keys(pruned.endpoints)) {
         if (
-          pruned.endpoints[endpointK as EndpointId]?.environments?.length === 0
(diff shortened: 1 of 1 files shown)


packages/fern-docs/bundle/src/server/docs-loader.ts
```diff
@@ -344,6 +345,20 @@ const createGetPrunedApiCached = (domain: string) =>
...
+        ) {
+          console.debug(
+            `${endpointK} has empty environments, adding default URL.`
+          );
+          pruned.endpoints[EndpointId(endpointK)]?.environments?.push({
+            id: "Default" as EnvironmentId,
+            baseUrl: "https://host.com",
+          });
+        }
+      }
```
dsinghvi, line 360: nit: we should move this up to line `340` and after so that we update the pruned nodes before they are stored in the cache

dsinghvi (approved):

PR #2336, opened 2025-03-16
Add fern-dashboard skeleton

zachkirsch:
Screenshot 2025-03-16 at 12 05 21 PM


zachkirsch:
Failing 3 CI checks but those are failing on main too so don't think related

dsinghvi (approved):
Sick! @zachkirsch would love to build this using app router + next.js and stay consistent with the new docs

I think that will allow us to embed WYSIWIG into the dashboard when the time is right. lmk if you disagree!

zachkirsch:
this is already using app router!

PR #2392, opened 2025-03-20
fix: use cache on routes

abvthecity:
https://github.com/fern-api/fern-platform/pull/2386/files triggers the following 500-level error:

```
Cannot access nextUrl on the server. You cannot dot into a temporary client reference from a server component. You can only pass the value through to the client.
```

This is because `.nextUrl`, `.searchParams`, and `cookies()` cannot be contained within a `"use cache"` function. This PR breaks it up.

### DIFF
--- a/packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/changelog/route.ts
+++ b/packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/changelog/route.ts
@@ -1,5 +1,6 @@
 import { unstable_cacheTag } from "next/cache";
 import { cookies } from "next/headers";
+import { notFound } from "next/navigation";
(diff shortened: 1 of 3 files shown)


packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/changelog/route.ts
```diff
@@ -22,29 +23,86 @@ export async function GET(
...
+  }
+}
+
+async function getJsonFeed(
+  host: string,
+  domain: string,
+  path: string,
+  fernToken: string | undefined
+): Promise<string> {
+  "use cache";
```
Comment (abvthecity) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/changelog/route.ts:54]: applying the "use cache" inside this function gets around the error

dsinghvi (approved):

PR #2402, opened 2025-03-21
chore: update meta parsing for wordwrap

chdeskur:
for composio

### DIFF
--- a/packages/fern-docs/bundle/src/mdx/plugins/rehype-code-block.test.ts
+++ b/packages/fern-docs/bundle/src/mdx/plugins/rehype-code-block.test.ts
@@ -65,4 +65,22 @@ describe("migrateMeta", () => {
       `" title="python.py" {metadata_that_means_nothing=true}"`
     );
   });
+
+  it("should remove wordWrap if it is next to the title", () => {
+    expect(migrateMeta("Python wordWrap maxLines=100")).toMatchInlineSnapshot(
+      `"title="Python"  wordWrap maxLines={100}"`
+    );
+  });
+
+  it("should remove wordWrap if that is the only word", () => {
+    expect(migrateMeta("wordWrap")).toMatchInlineSnapshot(
+      `"title="" wordWrap"`
+    );
+  });
+
+  it("should remove wordWrap if it is next to just a title", () => {
+    expect(migrateMeta("wordWrap myFile.txt")).toMatchInlineSnapshot(
+      `"wordWrap title="myFile.txt" "`
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


packages/fern-docs/bundle/src/mdx/plugins/rehype-code-block.test.ts
```diff
@@ -65,4 +65,22 @@ describe("migrateMeta", () => {
...
+    expect(migrateMeta("wordWrap")).toMatchInlineSnapshot(
+      `"title="" wordWrap"`
+    );
+  });
+
+  it("should remove wordWrap if it is next to just a title", () => {
+    expect(migrateMeta("wordWrap myFile.txt")).toMatchInlineSnapshot(
+      `"wordWrap title="myFile.txt" "`
+    );
+  });
```
dsinghvi, line 85: want to add a test for composio's actual usecase `json JSON maxLines=50 {14-19} {63-89} wordWrap`, I think it should work but probably a cheap thing to add

packages/fern-docs/bundle/src/mdx/plugins/rehype-code-block.ts
```diff
@@ -198,15 +202,20 @@ export function migrateMeta(metastring: string): string {
       if (text.trim() === "") {
         return "";
       }
-      return `title="${text.trim()}" `;
+
+      const lintText = text.replaceAll(/(wordWrap)/g, "").trim();
+      return text.replace(
+        lintText,
+        `title="${lintText.replace(/"/g, '\\"')}"${text.includes("wordWrap") ? " " : ""}`
+      );
```
dsinghvi, line 210: lines 192-196 seem pretty similar to 205-210, thoughts on a helper function here? 

```ts
if (...) {
  return createMetaWithTitleAttribute(metastring); 
}

if (...) {
  return createMetaWithTitleAttribute(text); 
}

function createMetaWithTitleAttribute(text: string): string {
      const textWithoutWordWrap = text.replaceAll(/(wordWrap)/g, "").trim();
      return text.replace(
        textWithoutWordWrap,
        `title="${textWithoutWordWrap.replace(/"/g, '\\"')}"${text.includes("wordWrap") ? " " : ""}`
      );
}
```

packages/fern-docs/bundle/src/mdx/plugins/rehype-code-block.test.ts
chdeskur: line 70 is from their use-case

chdeskur: where did you see this? the syntax for line highlighting is wrong, I think

dsinghvi (approved):

packages/fern-docs/bundle/src/mdx/plugins/rehype-code-block.test.ts
dsinghvi: https://github.com/ComposioHQ/composio/blob/3273162f38238904eb3f446a3d6ed1b35af99931/fern/tool-calling/processing-tools.mdx?plain=1#L15 ah interesting, just searched here

PR #2404, opened 2025-03-21
chore: handle hidden nodes

chdeskur:
this pr handles hidden inheritance, bringing over some of the changes that were made in https://github.com/fern-api/fern-platform/pull/2381


dsinghvi:
Would be great to make sure instabase and nominal continue to work.

packages/fern-docs/bundle/src/server/withPrunedNavigation.test.ts
```diff
@@ -0,0 +1,1125 @@
+import { NodeId, PageId, RoleId, Slug } from "@fern-api/fdr-sdk/navigation";
+
+import { withPrunedNavigation } from "./withPrunedNavigation";
+
+describe("withPrunedNavigation", () => {
+  it("child should be hidden if parent is hidden", () => {
+    expect(
+      withPrunedNavigation(
```
dsinghvi, line 8: nit: these tests could potentially be more readable with  some `createNode` helper functions. 

```ts
expect(withPrunedNavigation(createRootNode({ child: createSideBarRoot({ children: [createHidden()] }) })); 
```

dsinghvi: and then you can even use those nodes in the assertion itself

dsinghvi (approved):

PR #2408, opened 2025-03-21
chore: version matching

chdeskur:
adds version matching for app router


dsinghvi (approved):
wonder if there is a test we can run here

PR #2414, opened 2025-03-22
feat: versioned explorer sidebar

chdeskur:


chdeskur:
out of scope: using the version dropdown to switch between versions of the explorer

packages/fern-docs/bundle/src/components/playground/utils/flatten-apis.ts
```diff
@@ -15,13 +15,18 @@ const trimBreadcrumbs = createBreadcrumbSlicer<ApiGroup>({
...
-  root: FernNavigation.NavigationNode | undefined
+  root: FernNavigation.NavigationNode | undefined,
+  version: string | undefined
 ): ApiGroup[] {
   if (root == null) {
     return [];
   }
   const result: ApiGroup[] = [];
   FernNavigation.traverseDF(root, (node, parents) => {
+    if (version && node.type === "version" && node.versionId !== version) {
```
abvthecity, line 26: Can you move this upstream?

abvthecity (approved):

packages/fern-docs/bundle/src/components/playground/utils/flatten-apis.ts
chdeskur: could you explain a bit more on what you mean?

chdeskur: thank you! resolved

PR #2422, opened 2025-03-23
updating nextUrl caching:

sfc-gh-kwells:
Fixes FER-

500 Error when trying to call llms-full.txt page due to "use cache" was causing the `req.nextUrl` to be undefined.
Error: Cannot access nextUrl on the server. You cannot dot into a temporary client reference from a server component. You can only pass the value through to the client.
at <unknown> (.next/server/app/[host]/[domain]/api/fern-docs/llms-full.txt/route.js:1:4059)

Removed token to get page to load.
Locally tested 


packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/llms-full.txt/route.ts
```diff
@@ -41,10 +39,10 @@ async function getLlmsFullTxt(
 ): Promise<string> {
   "use cache";
 
-  unstable_cacheTag(domain, "llms-full-txt");
+  unstable_cacheTag(domain, "getLlmsFullTxt");
 
-  const fern_token = (await cookies()).get(COOKIE_FERN_TOKEN)?.value;
-  const loader = await createCachedDocsLoader(host, domain, fern_token);
+  // const fern_token = (await cookies()).get(COOKIE_FERN_TOKEN)?.value;
+  const loader = await createCachedDocsLoader(host, domain);
```
Comment (dsinghvi) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/llms-full.txt/route.ts:45]: I havent traced the code to see what the impact of not providing the fern toke is, i think its okay

dsinghvi (approved):

PR #2441, opened 2025-03-26
feat: twoslash alpha

chdeskur:
this pr sets up an alpha of the twoslash integration. 

in scope: 
- any codeblock with `twoslash` in the meta will be pre-processed by `shiki/twoslash`
- codeblocks are styled to match the UI theme

out of scope (for now): 
- codeblocks can include a title + other meta descriptors (e.g. line highlighting)

fast-follows: 
- twoslash docs hoverables will render markdown (instead of plaintext)

https://github.com/user-attachments/assets/91a13d88-4a69-4867-a2c5-bcfbbf90ddd4


package.json
```diff
@@ -49,7 +49,9 @@
     "micromatch": "4.0.8",
     "react": "19.0.0",
     "react-dom": "19.0.0",
+    "shiki": "3.2.1",
     "typescript": "5.7.3",
+    "unified": "11.0.5",
```
dsinghvi, line 54: any reason we had to add to `resolutions` instead of changing the package everywhere?

dsinghvi (approved):
lgtm, one minor comment

package.json
chdeskur: i think one of the shiki packages also used unified which is why this had been added at some point. seems to work fine when removed, but better safe than sorry?

chdeskur: seems like the convention we're using is to explicitly resolve certain packages we know will be used in other dependencies.

PR #2454, opened 2025-03-27
fix: add a fallback for the root

chdeskur:
this pr adds a fallback for calculating the endpoints that should appear in the explorer sidebar. 

we first search for the endpoints assigned to the current version. if we can't filter to a version (e.g. the site is unversioned), we just use the root as context.


chdeskur:
Screenshot 2025-03-26 at 11 04 13 PM
Screenshot 2025-03-26 at 11 03 50 PM
Screenshot 2025-03-26 at 11 01 18 PM

dsinghvi (approved):

PR #2465, opened 2025-03-28
fix: enable edit of environments on double click

chdeskur:
this pr does three things
- ensures the environment is always able to be edited by double clicking
- restructures the HTML not to nest two buttons within each other 
- fixes broken tooltips within the API explorer


chdeskur:
https://github.com/user-attachments/assets/44a64453-665e-4b98-b8d6-79e9aae68a15

kgowru (changes requested):
Can you decrease the border-radius around the URL tag in API explorer a few clicks? Currently looks like a pill.

dsinghvi (approved):

kgowru (approved):
lgtm

PR #2466, opened 2025-03-28
chore: placeholder update

chdeskur:
improves placeholder for ai-enhanced search

this pr also updates the searchbar button to use `not-allowed` cursor when search is loading (e.g. in local preview mode)

### DIFF
--- a/packages/fern-docs/bundle/src/state/search.tsx
+++ b/packages/fern-docs/bundle/src/state/search.tsx
@@ -4,13 +4,31 @@ import React from "react";
 
 import { composeEventHandlers } from "@radix-ui/primitive";
 import { atom, useAtomValue, useSetAtom } from "jotai";
+import { useHydrateAtoms } from "jotai/utils";
 
 import { DesktopSearchButton } from "@fern-docs/search-ui";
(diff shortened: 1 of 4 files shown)


chdeskur:
Screenshot 2025-03-28 at 12 44 20 AM

packages/fern-docs/bundle/src/state/search.tsx
```diff
@@ -43,6 +64,13 @@ export const SearchV2Trigger = React.memo(function SearchV2Trigger(
       {...props}
       onClick={composeEventHandlers(props.onClick, toggleSearchDialog)}
       variant={isInitialized ? "default" : "loading"}
+      placeholder={
+        isAskAiEnabled
+          ? isMobile
+            ? "Search or ask AI"
+            : "Search docs or ask AI a question ✨"
```
fern-support, line 71: Sorry, i used an emoji in Slack but i was intending it to be an icon. 

Upon further thought, let's just remove it and keep it simple. That would be too many icons.

chdeskur: updated!

fern-support (approved):

PR #2481, opened 2025-04-01
feat: have AI chat respect docs.yml config

dubwub:
Fixes FER-

* fdr-generated code to get latest fdr-sdk
* added modelid + system prompt logic to
TODO: once webflow re-pushes their docs.yml for webflow-ai, we can remove the last vestiges of custom code

We want to support custom workflows for different customers (e.g. what model, what system prompt)

Locally tested


dsinghvi (approved):
Let's make sure that COHERE_API_KEY is added to `prod.ferndocs.com` and `dev.ferndocs.com`

PR #2502, opened 2025-04-04
chore: add fallback for dark code

chdeskur:
this pr adds a fallback for the dark color theme if not specified in the user's docs.yml configuration. 

this handles an edge case where a site defines only light-mode but uses dark code blocks.


chdeskur:
before:
Screenshot 2025-04-04 at 1 09 32 AM

after:
Screenshot 2025-04-04 at 1 07 37 AM

dsinghvi (approved):

PR #2512, opened 2025-04-05
fix: disable playground examples in certain conditions

dsinghvi:
Allow certain customers to migrate to app router. 

Unblock app router. 

Preview URls.


chdeskur (approved):

dsinghvi:
> could you include some context on how we can eventually migrate off the domain-specific code? ideally as a short comment in the code + slightly more detail in the pr?

Options include: 
- Configuring in `docs.yml`
- Configure in edge config
- Improving OpenAPI parsing so that no configuration is needed (still unclear if this is possible)

Option 3 would be idea.

PR #2513, opened 2025-04-05
fix: multiform python snippets

chdeskur:
this PR fixes a few aspects of python snippets, specifically for multipart form data. 
- removes the multipart content header if files are sent
- remove unnecessary `json.dumps` (only required if the field is an actual json object)
- update the files parameter to accept a python dictionary instead of a list
- deconstruct an array parameter if it only has one value
- include exploded parameters


packages/fern-docs/bundle/src/components/playground/code-snippets/builders/python.ts
```diff
@@ -73,18 +73,27 @@ ${this.#buildRequests({ json: JSON.stringify(this.maybeWrapJsonBody(value), unde
...
-              : `'${k}': json.dumps(${indentAfter(JSON.stringify(v.value, undefined, 2), 2, 0)}),`
+              : `'${k}': ${
+                  typeof v.value === "object" && !Array.isArray(v.value)
+                    ? `json.dumps(${indentAfter(JSON.stringify(v.value, undefined, 2), 2, 0)})`
+                    : typeof v.value === "object" &&
+                        Array.isArray(v.value) &&
+                        v.value.length === 1
+                      ? JSON.stringify(v.value[0])
+                      : JSON.stringify(v.value)
+                },`
```
dsinghvi, line 96: would be great to extract this into a helper function along the lines of `getFormParameterValue` and use a bunch of if statements instead of ternary operators

packages/fern-docs/bundle/src/components/playground/code-snippets/builders/python.ts
```diff
@@ -94,7 +103,7 @@ ${this.#buildRequests({ json: JSON.stringify(this.maybeWrapJsonBody(value), unde
             ? `{\n${indentAfter(dataEntries, 2)}\n}`
             : undefined;
 
-        if (data != null) {
+        if (data?.includes("json.dumps")) {
```
dsinghvi, line 106: nice

dsinghvi (approved):
@chdeskur these look good to me!

PR #2514, opened 2025-04-06
fix: file upload in explorer

chdeskur:
this PR modifies file upload behavior in the explorer to fail more gracefully. 

when a user navigates away from the page, the file state is lost. in lieu of trying to store the file data another way, we should more definitively remove the file value so that we don't try to render an invalid file again. 

tried locally with ideogram API explorer (the site that reported the error to begin with)


dsinghvi (approved):
LGTM

PR #2527, opened 2025-04-07
chore: add wait for revalidate

chdeskur:
this PR attempts to resolve the occasional issue of images failing to use the latest version. 

we believe the problem comes from the revaliate endpoint. when revalidateTag is called, it should invalidate the cache, but in some logs, there is a cache hit after revalidation (this should not happen). the cache hit occurs when the cache is accessed before the revalidateTag call finishes executing. 

this appears to be an open issue in the app router: https://github.com/vercel/next.js/discussions/53206

we're implementing a small delay to avoid this issue. if the issue persists, we can investigate the double caching layer in the


packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/revalidate/route.ts
```diff
@@ -49,6 +49,9 @@ export async function GET(
   const { host, domain } = await props.params;
   revalidateTag(domain);
 
+  // delay to ensure invalidation propagates before cache is accessed
+  await new Promise((resolve) => setTimeout(resolve, 100));
```
Comment (dsinghvi) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/revalidate/route.ts:53]: do you know if `revalidateTag` is async and we should await it?

Comment (chdeskur) [reply packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/revalidate/route.ts:53]: would be so nice if it was. that's what people are complaining about in the linked vercel issue.

PR #2550, opened 2025-04-08
Add progress bar when changing route

zachkirsch:
Screenshot 2025-04-08 at 7 52 31 PM


packages/fern-dashboard/src/providers/ProgressProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+
+export function ProgressProvider({ children }: ProgressProvider.Props) {
+  const { resolvedTheme = "light" } = useTheme();
+
+  return (
+    <BProgressProvider
+      height="4px"
+      color={resolvedTheme === "light" ? "#008700" : "#70e155"}
```
kgowru, line 19: Can we use var(--green-6) instead of hardcode hex values?

kgowru (dismissed):

packages/fern-dashboard/src/providers/ProgressProvider.tsx
zachkirsch: ah cool didn't even consider that would work with their `color` prop but works great! good call

dsinghvi (approved):

PR #2560, opened 2025-04-09
Fix FDR archiving for docs sites w/o paths

zachkirsch:
In FDR's `setIsDocsDefinitionArchived`, of using `URL.pathname` for the pathname (which can sometimes be `/`), use `ParsedBaseUrl` to match how we are constructing pathnames elsewhere in the codebase


servers/fdr/src/db/docs/DocsV2Dao.ts
```diff
@@ -582,14 +582,14 @@ export class DocsV2DaoImpl implements DocsV2Dao {
...
-    url: URL;
+    url: ParsedBaseUrl;
     isArchived: boolean;
   }): Promise<void> {
     await this.prisma.docsV2.update({
       where: {
         domain_path: {
           domain: url.hostname,
-          path: url.pathname,
+          path: url.path ?? "",
```
zachkirsch, line 592: root cause of archiving failing for docs sites w/o paths is: `URL.pathname` was `/` but `ParsedBaseUrl.path` was `undefined`

PR #2607, opened 2025-04-16
chore: remove fern token, prepare to read from fdr cli-side

chdeskur:
this PR further reduces the dependency on the `.env.local` secrets by eliminating the need for an admin `fern_token`, `qstash` creds, and `kv` creds. 

this PR also updates some logic around disabling buttons:
- the login button no longer appears for every site
- the api explorer is now able to be opened, but sending a request is blocked

remaining work: 
- when using the locally hosted server, images do not render
- edge-config needs to be removed


packages/fern-docs/bundle/src/server/registry.ts
```diff
@@ -3,14 +3,27 @@ import { once } from "es-toolkit/function";
 import { FdrClient } from "@fern-api/fdr-sdk/client";
 
 import { fernToken_admin } from "./env-variables";
+import { isLocal } from "./isLocal";
 
 function getEnvironment() {
+  // todo: update this to point at the correct local enviroment
+  if (isLocal()) {
+    return "http://localhost:3001";
```
dsinghvi, line 11: @chdeskur this will need to be dynamic b/c the preview server can run on a dynamic port based on what port is available on the user's computer

chdeskur: yep, figured there's a bit more logic to be added here but wanted to scope to a different PR since this is getting a bit large

packages/fern-docs/bundle/src/server/docs-loader.ts
```diff
@@ -783,6 +913,14 @@ const getFonts = cache(async (domain: string) => {
 
   unstable_cacheTag(domain, "getFonts");
 
+  if (isLocal()) {
+    const response = await loadWithUrl(domain);
+    return generateFonts(
+      response.definition.config.typographyV2,
+      await getFiles(domain)
```
dsinghvi, line 920: what does `getFiles` do ? (may be related to images not rendering)

packages/fern-docs/bundle/src/server/docs-loader.ts
```diff
@@ -696,6 +759,73 @@ const getColors = cache(async (domain: string) => {
...
+              border: toOklch(dark.border),
+              sidebarBackground: toOklch(dark.sidebarBackground),
+              headerBackground: toOklch(dark.headerBackground),
+              cardBackground: toOklch(dark.cardBackground),
+            }),
+            backgroundGradient: dark.background.type === "gradient",
+          }
+        : undefined,
+    };
+  }
```
dsinghvi, line 827: weird, why do we have to do this for local? ideally local and production use the exact same logic here

packages/fern-docs/bundle/src/server/registry.ts
dsinghvi: yup lets remove entirely till we can figure that out

PR #2611, opened 2025-04-16
chore: remove edge config, local fdr call

chdeskur:
this PR removes reliance on the edge config, and points fdr to the local development host

remaining work: 
- fix files not showing up 
- confirm `host` approach in `loadWithUrl` once the CLI is set up


dannysheridan:
I didn't think I'd get so emotional. Congratulations

dsinghvi (approved):
🚀

PR #2613, opened 2025-04-17
chore: add notifications for s3 storage failures

dsinghvi:
S3 storage failures will drastically increase pressure on the backend and increase function duration. 

See how often s3 storage failure is happening. 

N/A


servers/fdr/src/services/s3/S3Service.ts
```diff
@@ -152,7 +155,22 @@ export class S3ServiceImpl implements S3Service {
...
+        error
+      );
+      // Send a slack notification about the failure
+      await this.app.services.slack.notify({
+        message: `Fail to store docs for ${domain} in s3!`,
+        err: error,
+      });
+
+      throw error;
+    }
```
dsinghvi, line 173: @chdeskur the important part

chdeskur (approved):

PR #2619, opened 2025-04-18
chore: respect list type

chdeskur:
this PR updates the ordered list component so that we respect the `type` attribute

also fixes a small error with the fern link component to pass href the correect empty type


chdeskur:
Screenshot 2025-04-18 at 9 40 57 AM

dsinghvi (approved):

PR #2649, opened 2025-04-26
chore: reduce bundle size

chdeskur:
this PR experiments with reducing the size of the local development bundle. 

it removes all folders and files other than the `standalone` folder. works as expected locally, but need to deploy to CDK to observe behavior in the build environment.


chdeskur:
this reduces the bundle size from 300 MB -> 50 MB

dsinghvi (approved):

PR #2652, opened 2025-04-26
chore: reinstall esbuild

chdeskur:
this PR includes a utility to be sent along with the local preview bundle to resolve the esbuild dependency on a user's machine


packages/cdk/utilities/install-esbuild/src/index.ts
```diff
@@ -0,0 +1,69 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as os from 'os';
+import { execSync } from 'child_process';
```
dsinghvi, line 4: nit: i prefer to use async/awaitable libraries than `sync`, if there's an awaitable version of child_process would recommend

packages/cdk/utilities/install-esbuild/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+const SKIP_DIRS = ['node_modules', '.git', 'dist', 'build'];
+
+async function processFiles(dir: string, platform?: string, arch?: string): Promise<void> {
+  const entries = fs.readdirSync(dir, { withFileTypes: true });
```
dsinghvi, line 9: nit: would recommend using `fs/promises`

packages/cdk/utilities/install-esbuild/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    
+    if (entry.isDirectory() && !SKIP_DIRS.includes(entry.name)) {
+      await processFiles(fullPath, platform, arch);
+    } else if (entry.isFile() && entry.name.endsWith('.json')) {
+      let content = fs.readFileSync(fullPath, 'utf8');
+      
+      if (content.includes('linux-x64')) {
+        const replacement = `${platform}-${arch}`;
+        content = content.replace(/linux-x64/g, replacement);
+        fs.writeFileSync(fullPath, content, 'utf8');
```
dsinghvi, line 22: same here, would rather do `await writeFile(...)`

dsinghvi (approved):

PR #2660, opened 2025-04-28
chore: always close the nav on tab change

chdeskur:
this PR updates the tab navigation behavior so that sections are always closed when navigating away from a tab

https://github.com/user-attachments/assets/cd28bb6a-1c67-42b9-a967-03ddeb79d1d3


sbawabe (approved):
lgtm!!

PR #2674, opened 2025-04-29
feat: openrpc playground actually works

dsinghvi:
If you go to https://alchemy.docs.buildwithfern.com/node/ethereum/ethereum-api-endpoints/eth-uninstall-filter/~explorer and try to use the API explorer, it doesnt really work. 

This PR introduces a `PlaygroundOpenRPCParameterForm.tsx` which handles OpenRPC's quirks in the API playground.


packages/fern-docs/bundle/src/components/playground/form/PlaygroundOpenRPCParameterForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+          const oldArray = castToArray(oldValue);
+          const newArray = [...oldArray];
+          newArray[index] =
+            typeof newValue === "function"
+              ? newValue(newArray[index])
+              : newValue;
+          return newArray;
+        });
+      },
+      [onChange]
```
dsinghvi, line 123: this is the key change -- we cant leverage the EndpointBodyForm bc it castsToRecord but OpenRPC parameters are sent as an array so values/index are critical

PR #2681, opened 2025-05-02
chore: enable search by breadcrumb

chdeskur:
this PR adds `breadcrumb` to the list of searchable attributes. 

searchable attributes should be kept limited to realistic queries a user might enter. one potential use-case is a docs site with numerous APIs. each API section may have an overview page, along with some endpoints. if a user knows they care only about API X, we should allow them to search for that API section, including both documentation and endpoints in the query. this is the justification for using breadcrumbs. 

before: 
Screenshot 2025-05-02 at 9 32 15 AM
after:
Screenshot 2025-05-02 at 9 30 16 AM


sbawabe (approved):
woooooof long hour dude

PR #2683, opened 2025-05-02
feat(ui): endpoint schema snippet

sbawabe:
DOCS-4

- removes excess `server-only`s and `use client`s to allow endpoint components to be reusable on both the client and server sides
- removes `serialize` prop drilling to ease usability on client side, and adds in `MdxSerializerContext` for server to share `serialize` function to deeply nested children
- loosen strictness of client vs server to maximize reusability and allow Next.js to optimize server vs. client

- `EndpointSchemaSnippet` didn't work yet in App router due to its requirements for once-server-only components
- in order to pipe this component over, we have to loosen the requirements of the once-server-only components to be unspecified. this will allow us to rely instead on Next.js to optimize SSR, in exchange for us being able to reuse those those now-loosened components on both the server and client sides.
- **note:** i did NOT loosen all components yet, i kept this PR within the scope of only what was required for the `EndpointSchemaSnippet`. there's likely more work to do here for other components!

Confirmed using payroc that their `EndpointSchemaSnippet` renders correctly!
Screenshot 2025-05-02 at 4 06 00 PM


packages/fern-docs/bundle/src/components/PageHeader.tsx
```diff
@@ -68,7 +68,7 @@ export function PageHeader({
             )}
             {tags}
           </div>
-          {includeDropdown && (
+          {includeDropdown && markdown && (
```
sbawabe, line 71: **note:** we were running into a lot of lint errors because `markdown` and `includeDropdown` became required props, but we didn't _actually_ add them to all PageHeader instances. marked these as optional and now we check both are truthy before we render the `PageActionsDropdown`!

dsinghvi: thank you!

packages/fern-docs/bundle/src/components/api-reference/endpoints/utils.ts
```diff
@@ -8,3 +10,23 @@ export function convertNameToAnchorPart(
...
+  if (maybeMethod != null) {
+    method = maybeMethod.toUpperCase() as HttpMethod;
+  }
+
+  // ensure that method is a valid HTTP method
+  if (method == null || !HttpMethod[method] || path == null) {
+    return undefined;
+  }
+
+  return { method, path };
```
dsinghvi, line 31: bonus: if there's some way to report or flag whenever someone has an endpoint that cant be resolved that would be great, but lets not block on building that

dsinghvi (approved):
@sbawabe this lgtm

PR #2688, opened 2025-05-04
Add github commands to cli

jsklan:
The following commands are being added: 
- `generator-cil github push --config <>`

Also added stubs for commands to be added soon:
- `generator-cil github pr --config <>`
- `generator-cil github release --config <>`

This will allow for self-hosted SDK generators. 

Self hosting github pushing in the SDK generators instead of going through fiddle. 

So far only locally testing against test repo: [test-generate-cli-github](https://github.com/fern-api/test-generate-cli-github)


packages/generator-cli/src/github/GitHub.ts
```diff
@@ -0,0 +1,49 @@
...
+      const wd = cwd();
+
+      const sourceDirectory = resolve(wd, this.githubConfig.sourceDirectory);
+
+      const repository = await cloneRepository({
+        githubRepository: this.githubConfig.uri,
+        installationToken: this.githubConfig.token,
+      });
+
+      const branch = this.githubConfig.branch ?? "main";
```
dsinghvi, line 28: nit: is there a way to dynamically get the default branch?

dsinghvi (approved):

packages/generator-cli/src/github/GitHub.ts
jsklan: fixed

PR #2698, opened 2025-05-06
fix(explorer): ensure explorer persists on page refresh

sbawabe:
Fixes DOCS-1

- Changes the `~explorer` route to instead be a query parameter (`explorer=true`) so that we persist modal state on page refresh, rather than leverage the intercepted route
- Maintains the parallel routing structure of the drawer so that the modal still renders on top of the page, rather than cause a full route transition

We want the drawer to re-open at variable height when a page is refreshed or redirected to -- most notably when authenticating and coming back to the drawer

- [x] Persists variable height open on page refresh
- [x] Passes URL param to redirect URI on authentication\
- [x] Works on dynamic and static routes (tested using authed docs host)


packages/fern-docs/bundle/src/components/playground/utils/explorer-route.ts
```diff
@@ -8,13 +8,19 @@ import {
...
  */
 export function conformExplorerRoute(slugOrPathname: string): string {
-  return conformTrailingSlash(addLeadingSlash(slugOrPathname) + "/~explorer");
+  return conformTrailingSlash(
+    addLeadingSlash(slugOrPathname) + "?explorer=true"
+  );
 }
 
 export function isExplorerRoute(pathname: string): boolean {
-  return removeTrailingSlash(pathname).endsWith("/~explorer");
```
chdeskur, line 15: is it worth leaving a redirect from `/~explorer` to `?explorer=true` so we don't break?

packages/fern-docs/bundle/src/middleware.ts
```diff
@@ -175,18 +175,6 @@ export const middleware: NextMiddleware = async (request) => {
     }
   }
 
-  /**
-   * Rewrite .../~explorer to /[domain]/explorer/...
-   */
```
chdeskur, line 180: perhaps we route from `/~explorer` to `?explorer=true`?

packages/fern-docs/bundle/src/components/playground/utils/explorer-route.ts
sbawabe: can do!

packages/fern-docs/bundle/src/middleware.ts
sbawabe: can do!

packages/fern-docs/bundle/src/components/playground/utils/explorer-route.ts
sbawabe: added!

packages/fern-docs/bundle/src/components/playground/PlaygroundCloseButton.tsx
```diff
@@ -23,7 +23,7 @@ export function PlaygroundCloseButton() {
         size="large"
         rounded
         variant="outlined"
-        href={pathname.replace("/~explorer", "")}
+        href={pathname.replace("explorer=true", "")}
```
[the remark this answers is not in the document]
sbawabe: my thinking here was that in case we had other query parameters, we want to make sure that we remove the one that's most relevant -- if we include removing the `?` here, then a pathname like: `/hello?hi=sup&explorer=true` would not replace & remove the `explorer=true` piece if we require it be the first and only query parameter.

i didn't want to mess too much with this button, but an ideal solution here would probably be to create a hook that gives us easy function access to remove and add query parameters, and then clicking this button would call some sort of `removeParam('explorer')` function that's provided to us by the hook. since this button was `href` based though, i didn't bother here to make a bigger change!

PR #2699, opened 2025-05-06
fix(github cli) Fixed issues with generator-cli github

jsklan:
Fixes FER-
- Shortened commit message: "SDK Generation"
- Added support for .fernignore, so files marked in .fernignore aren't deleted from remote

- Feedback from @dsinghvi based on initial usage

Tested locally against the [test repo](https://github.com/fern-api/test-generate-cli-github/blob/jsklan/add_fernignore/ignored.txt). Branch with examples is here.


packages/commons/github/src/ClonedRepository.ts
```diff
@@ -32,16 +34,39 @@ export class ClonedRepository {
...
+    ]);
+  }
+
+  public async restoreFiles(files: string[]): Promise<void> {
+    await this.git.cwd(this.clonePath);
+    await this.git.raw([
+      "restore",
+      ...(Array.isArray(files) ? files : [files]),
+    ]);
+  }
```
amckinney, line 61: I wonder if we should use an option here and consolidate these functions.

```suggestion
  public async restoreFiles({
    files,
    staged
  }: {
    files: string[],
    staged?: boolean
  }): Promise<void> {
    await this.git.cwd(this.clonePath);
    const args = ["restore"];
    if (staged) {
      args.push("--staged")
    }
    await this.git.raw([
      ...args,
      ...(Array.isArray(files) ? files : [files]),
    ]);
  }
```

That way it's simpler for us to add other options in the future without rolling out separate functions.

packages/commons/github/src/ClonedRepository.ts
```diff
@@ -32,16 +34,39 @@ export class ClonedRepository {
...
+
   public async commit(message?: string): Promise<void> {
+    const statusBeforeCommit = await this.git.status();
+    console.log("Status before commit:", statusBeforeCommit);
```
amckinney, line 65: Ideally we use a logger here, but I know that's difficult given that we don't have a class that wraps these utilities. Maybe we drop these for now? Otherwise we couldn't use this library without spilling logs into `stdout`.

amckinney: If we actually need these logs, I'd prefer adding an optional `logger` argument that the caller can thread in if they want.

packages/generator-cli/src/github/GitHub.ts
```diff
@@ -27,15 +28,18 @@ export class GitHub {
 
       const branch =
         this.githubConfig.branch ?? (await repository.getDefaultBranch());
+      console.log("Using branch:", branch);
```
amckinney, line 31: Given that this is in a class, this would be a good candidate for a proper logger.

amckinney (approved):
Discussed offline about incorporating `@fern-api/logger`. Feel free to add those in a follow-up if need be.

packages/commons/github/src/ClonedRepository.ts
jsklan: Oops, I meant to remove these. and yeah I agree with using the logger arg if we need them later.

packages/generator-cli/src/github/GitHub.ts
jsklan: removed

packages/commons/github/src/ClonedRepository.ts
jsklan: removed

PR #2710, opened 2025-05-07
chore: update node modules copy

chdeskur:
This PR updates the way we install the node modules required for the two-slash integration. It now installs all of them once the first time.


dsinghvi (approved):
semi blind +1

servers/mdx-bundler/src/serialize.ts
```diff
@@ -105,84 +105,110 @@ async function serializeTwoslashImpl(
...
+    "@account-kit/smart-contracts": "^4.31.2",
+    "@account-kit/react-native-signer": "^4.31.0",
+    "@tanstack/react-query": "^5.71.1",
+    google: "link:next/font/google",
+    "qrcode.react": "^4.2.0",
+    react: "^19.0.0",
+    "react-dom": "^19.0.0",
+    "react-native": "^0.79.2",
+    "ts-essentials": "^10.0.4",
+    viem: "2.22.6",
```
dsinghvi, line 124: can we just hardcoded these into our package.json?

dsinghvi: we can keep it dynamic though for future proofing and migrating everyone to this

PR #2729, opened 2025-05-14
feat(ci): Check if fern sites are up

davidkonigsberg:
- For all fern sites (according to vercel + the list of additional sites), check if the site is up
- If not, create an incident.io incident
- Also run in EU via self-hosted runner

We want to know about outages before customers notice them.

[Run in CI.](https://github.com/fern-api/fern-platform/actions/runs/15045254020/job/42286073143)


get_sites.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # Deal with pagination
+    next = response.json()["pagination"]["next"]
+    if next is None:    
+        done = True
+    else:
+        url = f"https://api.vercel.com/v10/projects/prod.ferndocs.com/domains?teamId={teamID}&limit=100&order=ASC&since={next}"
+
+# Skip domains in the list below
+skip_list = ["*.ferndocs.app","staging.ferndocs.com","twoslash.ferndocs.com","api.qdrant.tech","docs.wellesley.social","docs.skyflow.dev"]
```
davidkonigsberg, line 31: Worth confirming we want to skip all of these. Still unclean on why I'm seeing the flake. Requires further investigation.

davidkonigsberg: I figured this out. Needed to use Get not HEAD b/c of how the redirects work on those specific sites. Changing my wget flags to not use --spider but still avoid saving files fixed things.

dsinghvi (approved):

davidkonigsberg:
To do this in EU/other regions we will need to either proxy the requests through an EU server, setup an EU runner, or connect to a VPN. All options have their pros/cons. Running through a proxy is simple but requires a reliable proxy. A self-hosted runner would be nice, and if we do want to run these checks frequently, it might be cost effective... If we can get away with a t2.micro that's only like $9/month....

dsinghvi (approved):

dsinghvi:
> To do this in EU/other regions we will need to either proxy the requests through an EU server, setup an EU runner, or connect to a VPN. All options have their pros/cons. Running through a proxy is simple but requires a reliable proxy. A self-hosted runner would be nice, and if we do want to run these checks frequently, it might be cost effective... If we can get away with a t2.micro that's only like $9/month....

Cool, cost isn't a factor here -- lets set that up

PR #2733, opened 2025-05-14
fix(seo): ensure template title persists to child routes

sbawabe:
Fixes DOCS-44

- fixes SEO issue where `config.title` was not being persisted as page title suffix
- moves SEO `generateMetadata` functions to unified `seo.ts` file so that we centralize that logic into one place

- fixes regression

- ensure `| {config.title}` is maintained on all child routes
- confirmed that `sitemap.xml` files are identical for this review app & a major customer's `sitemap.xml`


packages/fern-docs/bundle/src/app/[host]/[domain]/dynamic/@explorer/[slug]/page.tsx
```diff
@@ -55,25 +54,3 @@ export default async function ExplorerPage({
 
   return <ExplorerContent loader={loader} node={node} />;
 }
-
-export async function generateMetadata({
```
Comment (sbawabe) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/dynamic/@explorer/[slug]/page.tsx:59]: parallel routes should not have metadata, they will be treated as the terminal node and will always show their title as a value

dsinghvi (approved):

PR #2739, opened 2025-05-15
fix(ci): Check site twice before opening an incident

davidkonigsberg:
Site must fail a second check before being considered down.
Added a timeout (the false positive came from a very slow run, potentially indicative of runner issues)
Flag action as failing if any sites are down

Got a false positive from [this](https://github.com/fern-api/fern-platform/actions/runs/15051198059/job/42306180040) CI run.

Ran in test mode [here](https://github.com/fern-api/fern-platform/actions/runs/15051798261/job/42308145576).


davidkonigsberg:
If this continues to have issues and they remain on the US runner, it may be worth creating a self-hosted runner for US checks as well. Minimal cost and will likely have a more robust network connection than the GH hosted runners.

dsinghvi (approved):

PR #2770, opened 2025-05-20
chore: update naming convention

chdeskur:
this PR udpates the naming convention for the explorer preview response selector to match fern branding


sbawabe (approved):
TY!!

PR #2771, opened 2025-05-20
chore: remove explorer SEO

chdeskur:
this PR disallows any url with a query parameter in `robots.txt` to make sure the explorer isn't indexed. it also removes the duplicated `/` from the sitemap URL


sbawabe (approved):
lgtm!!

PR #2777, opened 2025-05-21
feat: add local Postgres to Dockerfile.self_hosted

ariel-langer:

Made some changes to get Postgres running locally inside the container that's built from `Dockerfile.self_hosted`.

We are making incremental changes to this `Dockerfile.self_hosted` file so that eventually customers can spin up a container that they can use for self-hosting their docs. One piece of that is getting postgres running locally inside the container.

1. built the new image: `docker build -f servers/fdr/Dockerfile.self_hosted -t fern-self-hosted .`
2. ran the image in a container and entered a shell inside the container: `docker run -it -e RUN_MODE=shell fern-self-hosted`
3. confirmed i could connected to our postgres db inside the container by running `psql -h localhost -d postgres`

Screenshot 2025-05-21 at 10 42 13 AM


servers/self-hosted/package.json
```diff
@@ -0,0 +1,13 @@
...
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "docker:self-hosted": "sh create_self_hosted_docker.sh",
+    "test:self-hosted": "vitest src/__test__/self-hosted --globals --config src/__test__/self-hosted/vitest.config.ts"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
+}
```
ariel-langer, line 13: @dsinghvi all the dependencies that i would need in the `/self-hosted` folder are already dependencies in the top level `fern-platform` directory and get installed there. should i also add them here or is that not necessary?

dsinghvi: i think you'll want to add them here as well

dsinghvi (approved):

PR #2790, opened 2025-05-22
feat: get MinIO running for self-hosted docs

ariel-langer:

Made some changes to get MinIO running locally inside the container that's built from Dockerfile.self_hosted

We are making incremental changes to this Dockerfile.self_hosted file so that eventually customers can spin up a container that they can use for self-hosting their docs. One piece of that is getting MinIO running locally inside the container.

Yes i confirmed that MinIO is running successfully on the container by following these steps:
1. built the image with  `sh create_self_hosted_docker.sh`
2. built the container with the MinIO ports exposed to my host machine: `docker run -p 9000:9000 -p 9001:9001 fern-self-hosted:latest`
3. Visited http://localhost:9001/ where the MinIO web UI should be
4. Confirmed that the MinIO Web UI was live + can login with user/password (minioadmin, minioadmin) + can insert data into the object database

Screenshot 2025-05-22 at 4 39 30 PM

Screenshot 2025-05-22 at 4 48 57 PM


servers/self-hosted/Dockerfile.self_hosted
```diff
@@ -19,11 +19,28 @@ EXPOSE 5432
...
+RUN chmod +x minio
+RUN mv minio /usr/local/bin/
+
+RUN addgroup -S minio-user && adduser -S minio-user -G minio-user && \
+    mkdir -p /data && chown -R minio-user:minio-user /data
+
+ENV MINIO_VOLUMES="/data" \
+    MINIO_SERVER_URL="http://localhost:9000" \
+    MINIO_ROOT_USER="minioadmin" \
+    MINIO_ROOT_PASSWORD="minioadmin"
```
ariel-langer, line 34: @dsinghvi i _think_ its fine to just leave these hardcoded as is for now in the Dockerfile because MinIO is running locally so this isn't actually exposing sensitive information but then i wonder if we should also run the server like this (in the run_self_hosted.sh script) so that it cannot be exposed to the host machine: 

`exec minio server ${MINIO_VOLUMES} --console-address "localhost:9001" --address "localhost:9000"`

dsinghvi: agreed

dsinghvi (approved):

PR #2798, opened 2025-05-24
chore: allow absolute images as icons

chdeskur:
Screenshot 2025-05-24 at 12 05 37 PM

ex: 
```
tabs: 
  docs: 
    display-name: Docs
    icon: fa-solid fa-globe
  api: 
    icon: <img src="https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQmMCCPvUcDqOMpnwfVUd9FNUeJ9NbPC-tUGA&s" />
    display-name: API
```


dsinghvi (approved):
do we need to deal with sizing the icon at all since i imagine that'll be different across absolute paths?

PR #2800, opened 2025-05-27
feat: self hosted FDR

ariel-langer:

Made some changes to get FDR's server running locally inside the container that's built from Dockerfile.self_hosted; **this PR is just to get FDR's server running, not all endpoints will work if we hit them at the moment due to other bugs with self-hosting. I plan to fix these bugs in a later PR**

We are making incremental changes to our `fern-platform` codebase so that eventually customers can spin up a container that they can use for self-hosting their docs. One piece of that is getting FDR running locally inside the container. 

- wrote tests to confirm that `/health` check endpoint for FDR is returning 200 + wrote tests to confirm the DB migrations ran on the local postgres table
- i also checked these things manually on a docker container i spun up (`docker run -p 9000:9000 -p 9001:9001 -p 8080:8080 -it -e RUN_MODE=shell fern-self-hosted:latest`)

Screenshot 2025-05-27 at 11 49 15 AM
Screenshot 2025-05-27 at 11 48 54 AM


dsinghvi (approved):
Nice 👏 very easy to follow

PR #2802, opened 2025-05-27
feat: self hosted CI/CD tests

ariel-langer:

Currently the tests for self-hosted get skipped in our other workflows. We should run these tests on PRs to catch bugs that might break self-hosting. In this PR i add a CI/CD github workflow for running the tests for self-hosting.

To prevent bugs that might break self-hosting.

Putting up the PR runs the workflow and this PR can't get merged unless that workflow passes :) 

Screenshot 2025-05-27 at 2 37 52 PM


.github/workflows/run-self-hosted-tests.yml
```diff
@@ -0,0 +1,21 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - app
+  pull_request:
```
ariel-langer, line 7: @dsinghvi let me know if this looks right to you or we need to run this on anything else

turbo.json
```diff
@@ -36,6 +36,9 @@
     "test-ete": {
       "dependsOn": ["compile"]
     },
+    "test:self-hosted": {
+      "dependsOn": []
```
ariel-langer: wasn't **too** sure why the tests above use "compile", seems like i can run `pnpm turbo build test:self-hosted --filter=@fern-platform/self-hosted` fine without it so removed

turbo.json
```diff
@@ -36,6 +36,7 @@
     "test-ete": {
       "dependsOn": ["compile"]
     },
+    "test:self-hosted": {},
```
ariel-langer, line 39: wasn't too sure why the tests above use "compile", seems like i can run pnpm turbo build test:self-hosted --filter=@fern-platform/self-hosted fine without it so removed

PR #2806, opened 2025-05-27
feat: readonly/writeonly prop drilling

sahil485:


packages/fern-docs/bundle/src/components/api-reference/type-definitions/ObjectProperty.tsx
```diff
@@ -49,13 +52,15 @@ export const PropertyWithShape = React.memo(function PropertyWithShape({
...
   types,
+  location,
 }: {
   icon?: React.ReactNode;
   name?: string;
   description: string | undefined;
   availability: ApiDefinition.Availability | null | undefined;
   shape: ApiDefinition.TypeShape;
   types: Record<string, ApiDefinition.TypeDefinition>;
+  location?: "request" | "response";
```
dsinghvi, line 63: nit: can we make "request" | "response" an exported type and then reuse it everywhere

dsinghvi (approved):

PR #2816, opened 2025-05-28
add environment variable

chdeskur:
use environment variable to reference mdx-bundler service


dsinghvi (approved):
epic! would be great to then add this as a vercel env!

PR #2825, opened 2025-05-30
chore: add timeout for twoslash fallback

chdeskur:
adds a time limit for twoslash processing so that serialization doesn't fail with a timeout

logic to fallback to regular code snippets is already present


packages/fern-docs/bundle/src/mdx/bundler/serialize.ts
```diff
@@ -142,62 +145,82 @@ async function serializeMdxImpl(
     console.log(`Found ${twoslashBlocks.length} twoslash blocks to process`);
 
     if (twoslashBlocks.length > 0) {
```
dsinghvi, line 147: nit: can we move all this code into a helper function since the increased nesting makes it a little hard to read

chdeskur: fantastic idea

PR #2842, opened 2025-06-02
feat: Add http HEAD support for snippets

Swimburger:
Fixes FER-

Add HTTP HEAD support for snippets

We need HTTP head support in the SDKs, and the snippets produced need to go through FDR snippets.


packages/template-resolver/package.json
```diff
@@ -28,7 +28,7 @@
     "test": "vitest --run --passWithNoTests"
   },
   "dependencies": {
-    "@fern-fern/fdr-cjs-sdk": "0.76.19-1-6e46ece2a09",
```
Swimburger, line 31: @amckinney I regen'd using Fern and it gave me 0.0.6431.
Not sure if I did this right.

amckinney (approved):

PR #2846, opened 2025-06-02
fix: [DO-NOT-MERGE] [AI chat] remove docs from system-prompt, rely on tool calls instead

dubwub:
When injecting documents into system prompt, we are unable to get the URL metadata on the front-end. Relying on tool-calls (1) allows the LLM to make better query wording decisions and (2) allows us to get the beautified citation format in every message

Tested locally


chdeskur (approved):

dubwub:
Not merging this, seeing some weird behavior

PR #2858, opened 2025-06-03
feat: self-hosted NextJS app

ariel-langer:

We are trying to get the Next JS app running in our docker container on port 3000 for self-hosted docs.

This is one of the last pieces for an MVP of self-hosted docs.

This was tested by actually building + running the docker container. Confirmed that you can hit the local Next App and load the docs that you mounted the container on. These docs aren't rendering correctly but i'll figure that out in a follow up PR.

Screenshot 2025-06-06 at 5 00 29 PM


servers/self-hosted/Dockerfile.self_hosted
```diff
@@ -94,11 +95,22 @@ COPY --from=installer /app .
...
+# Setup Nextapp dependencies
+RUN mkdir -p /app/nextapp/
+COPY servers/self-hosted/docs_bundle.zip /app/servers/self-hosted/docs_bundle.zip
+RUN unzip -q /app/servers/self-hosted/docs_bundle.zip -d /app/nextapp/ && \
+    cd /app/nextapp/standalone/packages/fern-docs/bundle && \
+    pnpm install
+
+# Install esbuild (TODO investigate why this is needed)
+RUN cd /app/nextapp/standalone && \
+    pnpm add -w esbuild
```
ariel-langer, line 111: @dsinghvi its unclear to me why i needed this but was running into issues without it:

Screenshot 2025-06-04 at 1 41 32 PM

chdeskur: this is likely because we remove the esbuild node modules when building the local dev standalone bundle

servers/self-hosted/src/__test__/self-hosted/selfHostedDocs.test.ts
```diff
@@ -122,3 +122,23 @@ describe("FDR server is running and api endpoints are available", () => {
     expect(curlOutput).toBe("200");
   });
 });
+
+/** describe("Self-hosted docs has a running nextapp instance", () => {
```
ariel-langer, line 126: could not get this to work, i feel comfortable shipping without it for now though

PR #2863, opened 2025-06-04
chore: use hash in log

chdeskur:
for debugging


fern-support:
THIS IS CATHERINE - approving with fern support due to size of PR to unblock

PR #2867, opened 2025-06-04
fix: use dynamic org_name for self-hosted docs

ariel-langer:

We should have the org-name be dynamic in this code instead of hard coded in a .env file.

To clean up previous PR.

There's a test that covers this. I also changed the org name in one of my fern folders to "ariel-test" and then ran the docker image and confirmed we handled it correctly:

Screenshot 2025-06-04 at 2 41 21 PM
Screenshot 2025-06-04 at 2 41 09 PM


servers/self-hosted/.env
```diff
@@ -1,13 +1,10 @@
...
 
-# should match the org name in your fern folder
-ORG_NAME=example-org
-
 # S3/MinIO Configuration
 MINIO_USERNAME=minioadmin
 MINIO_PASSWORD=minioadmin
 MINIO_URL=http://localhost:9000
-MINIO_BUCKET_NAME=$ORG_NAME.docs.buildwithfern.com
+MINIO_BUCKET_NAME_SUFFIX=docs.buildwithfern.com
```
ariel-langer, line 7: @dsinghvi for customers that use a custom domain, should this bucket name match their custom domain instead? does this matter for an MVP for now?

PR #2895, opened 2025-06-09
chore: remove state-based logic for hiding the asides

chdeskur:
we should deterministically know when to hide the sidebar on a page. this PR removes the `<HideAsides />` logic and should increase stability of sidebar visibility.


packages/fern-docs/bundle/src/app/[host]/[domain]/dynamic/@sidebar/[slug]/page.tsx
```diff
@@ -91,7 +92,9 @@ export default async function SidebarPage({
           <SidebarTabsList tabs={found.tabs} />
         </SidebarTabsRootServer>
       )}
-      {!isSingleOverviewPage && (
+      {isSingleOverviewPage ? (
+        <HiddenSidebar />
```
Comment (chdeskur) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/dynamic/@sidebar/[slug]/page.tsx:96]: the calculation for an "empty" sidebar is done in the code above here, so there should be no need to include the <setemptysidebar /> state in the sidebar root. we want the styling to get injected still, so including the hidden sidebar styling here instead

packages/fern-docs/bundle/src/components/NotFoundContent.tsx
```diff
@@ -2,14 +2,15 @@ import "server-only";
...
+import { HiddenSidebar } from "@/state/layout";
 
 import ReturnHomeButton from "./ReturnHomeButton";
 
+// todo: don't hide the sidebar if disable-header is true
 export default async function NotFoundContent() {
   return (
     <>
-      <HideAsides force />
+      <HiddenSidebar />
```
chdeskur, line 13: this is fine for now, since it doesn't change any current functionality. todo is scoped to a separate pr

packages/fern-docs/bundle/src/components/sidebar/nodes/SidebarRootNode.tsx
```diff
@@ -75,15 +74,6 @@ export async function SidebarRootNode({
 
   return (
     <>
```
chdeskur, line 76: this computation is done earlier, and could have been the culprit for hydration errors

packages/fern-docs/bundle/src/components/layouts/AsideAwareDiv.tsx
```diff
@@ -2,18 +2,19 @@
 
 import React from "react";
```
chdeskur, line 3: `AsideAwareDiv` is now _only_ for changelogs, since the styling based on `data-aside-state` was only being applied to changelog pages anyways

dsinghvi (approved):

PR #2932, opened 2025-06-11
fix: add ws dependency override

ariel-langer:
We have a dependency on `ws: 7.5.9` which is a security vulnerability and need to upgrade to `7.5.10`. Not sure this is actually the correct approach to just override the version in our package.json file. Putting this up as a draft for now

To fix security vulnerabilities.

[TODO]

### DIFF
--- a/package.json
+++ b/package.json
@@ -151,7 +151,10 @@
       "serverless",
       "sharp",
       "workerd"
-    ]
+    ],
+    "overrides": {
+      "ws": "7.5.10"
+    }
   },
   "nextBundleAnalysis": {
     "buildOutputDirectory": "packages/fern-docs/bundle/.next",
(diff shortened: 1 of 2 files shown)


ariel-langer (commented):
just a WIP for now, there might be a better way to do this by upgrading another package. this is where we are installing the old version of ws in our pnpm-lock.yaml file:

```
 webpack-bundle-analyzer@4.10.1(bufferutil@4.0.9)(utf-8-validate@5.0.10):
    dependencies:
      '@discoveryjs/json-ext': 0.5.7
      acorn: 8.14.1
      acorn-walk: 8.3.4
      commander: 7.2.0
      debounce: 1.2.1
      escape-string-regexp: 4.0.0
      gzip-size: 6.0.0
      html-escaper: 2.0.2
      is-plain-object: 5.0.0
      opener: 1.5.2
      picocolors: 1.1.1
      sirv: 2.0.4
      ws: 7.5.9(bufferutil@4.0.9)(utf-8-validate@5.0.10)
    transitivePeerDependencies:
      - bufferutil
      - utf-8-validate
```

PR #2948, opened 2025-06-13
fix: fix workos secret vulnerability

ariel-langer:

Our self-hosted Docker container's bundled Next.js app code was being flagged by security scanning tools for exposing the WORKOS_API_KEY—even when the value was an empty string and even though the workos code doesn't actually get used by our self-hosted next js app. This issue stems from how the app is bundled at build time: the WorkOS client code was being included and flagged, despite the secret not actually being used for the self hosted version of the app.

To avoid these false positives, I've updated the build process for self-hosted docs to replace any import of @workos-inc/node with ./src/server/workos-stub.ts. This change ensures the actual WorkOS client is not bundled into our self-hosted Docker image, which is appropriate since WorkOS is not required for self-hosted deployments anyways.

This code is causing a security vulnerability in our self hosted docker container (trivy image fern-self-hosted:latest)
Screenshot 2025-06-11 at 4 41 44 PM

Rebuilt docker container + re-ran trivy image fern-self-hosted:latest and the security vulnerability is gone.


dsinghvi (approved):
👏 nice

PR #2953, opened 2025-06-13
feat: workflow for security vulnerabilities

ariel-langer:
Should check our self hosted docker container for security vulnerabilities.

Should check this to make sure we aren't introducing vulnerabilities

By running the workflow on github actions. Currently failing as expected because we still have a security vulnerability on cross-spawn package

Screenshot 2025-06-13 at 3 55 20 PM


ariel-langer (commented):
@dsinghvi i thought about adding this as a step to the "Publish Self-Hostable Docs" so we actually can't publish a new docker image if there's a security vulnerability but decided that would probably just unnecessarily block us from pushing updates to that container so i made this a separate workflow. i could be convinced that it should all be one workflow for now though (although we'd have to wait until the final critical dependency vulnerability is solved until we merge these into 1 workflow)

PR #2957, opened 2025-06-14
fix: get rid of stringified blob response

dsinghvi:
Occasionally we will see responses in the API playground be blob urls `blog:....` when they should be JSON. 
This updates the `executeProxyRest` logic to read content types for blobs, and then default to parsing as json (falling back to text). 

Customer complaints + prospect demos

- Tested on Voice AI customers for media download
- Tested on other customers that reported seeing an unexpected blob

### DIFF
--- a/packages/fern-docs/bundle/src/components/playground/fetch-utils/executeProxyRest.ts
+++ b/packages/fern-docs/bundle/src/components/playground/fetch-utils/executeProxyRest.ts
@@ -16,7 +16,6 @@ export async function executeProxyRest(
     Object.keys(req.headers).join(",")
   );
 
-  // multipart/form-data will be handled by the fetch API with a boundary, and should not be forwarded
   if (req.body?.type === "form-data") {
     requestHeaders.delete("Content-Type");
   }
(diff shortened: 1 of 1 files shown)


packages/fern-docs/bundle/src/components/playground/fetch-utils/executeProxyRest.ts
```diff
@@ -45,91 +44,91 @@ export async function executeProxyRest(
...
+        contentType: contentType || "text/plain",
+        time,
+        size,
+      };
     }
+  } catch {
+    // If text() fails, try as blob
+    const blob = await res.blob();
+    const endTime = Date.now();
+    return {
```
chdeskur, line 123: is there a case when blob would fail?

chdeskur (approved):

packages/fern-docs/bundle/src/components/playground/fetch-utils/executeProxyRest.ts
dsinghvi: from what i understand -- there isn't since everything can be turned into a blob

PR #2991, opened 2025-06-18
feat: add NullableType to FDR read/register

sahil485:
- add `NullableType` to FDR read/register
- `nullable` properties shouldn't be passed as `optional` in `convertIrtoFdrApi`
- manual


dsinghvi (approved):
👏

PR #3036, opened 2025-06-25
feat(askfern): Define initial query schemas.

eyw520:
This PR will define our Question/Answer db schema & set up local/dev/prod testing scripts.


servers/fai/tests/test_query_crud.py
```diff
@@ -0,0 +1,53 @@
...
+        "time_to_first_token": 0.5,
+    }
+
+    print(f"Sending POST request to create query with data: {query_data}")
+    response = requests.post(f"http://{docker_ip}:8080/queries", json=query_data, timeout=5)
+    print(f"Received response with status code: {response.status_code}")
+    assert response.status_code == 200
+    data = response.json()
+    print(f"Response data: {data}")
+    assert data["domain"] == query_data["domain"]
```
dsinghvi, line 24: lmao -- it would be nice to have an sdk here, this will work for now but for anything complex we're going to have a lot of issues

dsinghvi (approved):

PR #3038, opened 2025-06-26
chore: [Ask Fern] make default model claude sonnet 4

dubwub:
Tested locally with Elevenlabs

Goal is to get all customers on latest and greatest

Screenshot 2025-06-26 at 10 47 24 AM


dannysheridan:
Stoked to see this live for customers!

PR #3046, opened 2025-06-27
fix(ui) : missing hover state on errors in the API Reference

SaarthakMaini:
### **What was the issue?**

When hovering over error status codes (such as 4xx or 5xx) in the error example dropdown on API reference pages, there was **no visible hover effect**. In contrast, hovering over the “Success” (2xx) status code correctly showed a green background. This led to a confusing and inconsistent user experience, as error codes did not visually respond to hover like success codes did.

---

### **How did I investigate and find the root cause?**

1. **Component Analysis:**  
   I traced the dropdown rendering to the `FernSelectItem` component in `ErrorExampleSelect.tsx`. This component uses Radix UI’s Select and applies background colors on hover using the `data-[highlighted]` attribute and intent-based classes.

2. **Intent Mapping:**  
   The hover color is determined by the `intent` prop, which is set based on the HTTP status code using the `statusCodeToIntent` function.  
   - For 2xx codes, the intent is `"success"`, which maps to a green hover.
   - For 4xx/5xx codes, the intent was `"error"` (as set by `statusCodeToIntent`).

3. **CSS Class Check:**  
   The code for the hover effect was:
   ```tsx
   "data-[highlighted]:bg-(color:--red-a3)": intent === "danger",
   ```
   This means only items with `intent === "danger"` would get the red hover effect. However, error codes were mapped to `"error"`, not `"danger"`, so the red hover was never applied.

4. **Manual Test:**  
   When I hardcoded `intent="danger"`, the red hover effect worked, confirming that the intent mapping was the issue.

---

### **How did I solve it?**

- I updated the class assignment in `FernSelectItem` to apply the red hover effect for both `"error"` and `"danger"` intents:
  ```tsx
  "data-[highlighted]:bg-(color:--red-a3)": intent === "error" || intent === "danger",
  ```
- This ensures that regardless of whether the intent is set to `"error"` (from the status code mapping) or `"danger"` (from other sources), the red hover effect will be applied.

---

https://github.com/fern-api/fern/issues/7206

https://github.com/user-attachments/assets/5ce86c32-4d6f-433e-8882-42d3d9e801dc


chdeskur:
Thanks for this! Curious how you set up the testing environment?

SaarthakMaini:
Hello @chdeskur,

Just wanted to quickly follow up regarding the PR.

Thank you

Regards,
Saarthak

SaarthakMaini:
Hello @dsinghvi ,

Just wanted to request a quick review

Thanks

PR #3047, opened 2025-06-27
chore: build update, local file hosting for self-hosted

chdeskur:
Screenshot 2025-06-27 at 11 42 42 AM

use local file hosting for self-hosted docs

also updates the build process to unlink .env file


dsinghvi (approved):
that was easy 💥

PR #3062, opened 2025-07-01
feat: request write github scopes for visual editor

ariel-langer:
For visual editor, users will be able to open branches and commit changes against their docs sites. They will need to give Fern write permissions to their github account do this so that:

1. we can open branches in their name / against their repo
2. we can attribute commits to that user

Today users log in to Fern platform with their Github but we only request minimal permissions to read their email. We need to request additional permissions from the user to read/write to their repos. Previously as a hack, we were using a hardcoded GITHUB_TOKEN that belonged to our Fern support account.

Tested E2E. Confirmed that i can make branches + commit with my user's github token. Screenshot of PR that was opened with commits attributed to me instead of Fern support.

Before:
[TODO]

After:
Screenshot 2025-07-01 at 11 05 54 AM


packages/fern-dashboard/src/components/docs-page/DocsSiteOverviewCard.tsx
```diff
@@ -57,9 +56,83 @@ export function DocsSiteOverviewCard({
       branch: branchName,
       baseBranch: "main",
     });
-    console.log("response", response);
-    redirect(`/${orgName}/editor/${docsUrl}/${branchName}/root`);
-  }, [sourceRepo, session.user.name, orgName, docsUrl]);
+    if (response.success === false) {
+      toast.error("Failed to create branch");
```
ariel-langer, line 60: figured i would add a toast for this so we don't just fail silently

packages/fern-dashboard/src/components/docs-page/DocsSiteOverviewCard.tsx
```diff
@@ -57,9 +56,83 @@ export function DocsSiteOverviewCard({
       branch: branchName,
       baseBranch: "main",
     });
-    console.log("response", response);
-    redirect(`/${orgName}/editor/${docsUrl}/${branchName}/root`);
```
ariel-langer, line 61: @sbawabe this redirect wasn't working for me btw; ended up refactoring it

packages/fern-dashboard/src/components/docs-page/DocsSiteOverviewCard.tsx
sbawabe: perfect -- thank you!

PR #3066, opened 2025-07-01
feat(fdr): add type for grpc endpoints

sahil485:
- add `grpc: GrpcProtocol` mapping to endpoint `protocol` field
- add `GrpcProtocol` type (identical to `OpenRpcProtocol` type for now)

- custom UI for gRPC endpoints


fern/apis/fdr/definition/api/latest/endpoint.yml
```diff
@@ -52,6 +53,10 @@ types:
     properties:
       methodName: string
 
+  GrpcProtocol:
+    properties:
+      methodName: string
```
dsinghvi, line 58: you will need to add to v1 read and write as well and the migrators

dsinghvi (approved):

fern/apis/fdr/definition/api/latest/endpoint.yml
sahil485: Done!

PR #3067, opened 2025-07-02
Changelog RSS Button 

ariel-serruya:
I added a new component `RSSFeedButton` to `PageHeader` that gets rendered in the changelog. It leverages both `CopyToClipboardButton` and `FernButton`. It copies the RSS URL to clipboard (e.x. https://elevenlabs.io/docs/changelog.rss) on button click. I minimally customized the button's styles to more closely resemble the Figma wireframes (i.e. manually setting `gap-2`). 

You can find the original linear task [here](https://linear.app/buildwithfern/issue/FER-4107/add-an-rss-button-to-our-changelog). 
The latest Figma wireframes are found [here](https://www.figma.com/design/XP4AQDBA1YM5VCzKqGIJRm/Changelog-RSS-Button?node-id=1-164&t=BS2uZv3aL7r9gnYX-0)

https://github.com/user-attachments/assets/5e95889a-160a-495a-bdad-fb1fe8554514


packages/fern-docs/bundle/src/components/PageHeader.tsx
```diff
@@ -73,6 +76,7 @@ export function PageHeader({
               <PageActionsDropdown markdown={markdown} />
             </div>
           )}
+          {showRssFeedButton && <RSSFeedButton />}
```
chdeskur, line 79: any reason not to match the div styling used for the page actions? how does this look on mobile?

packages/fern-docs/bundle/src/components/RSSFeedButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+import { Rss } from "lucide-react";
+
+import { CopyToClipboardButton } from "@fern-docs/components/CopyToClipboardButton";
+import { FernButton } from "@fern-docs/components/FernButton";
+
+export function RSSFeedButton() {
+  const getRssUrl = () => {
+    const currentUrl = window.location.href;
+    const changelogUrl = currentUrl.split("/changelog")[0] + "/changelog";
```
chdeskur, line 13: it's not a given that the url will be called `changelog`. the user can set the page to an arbitrary slug, so i think you just want to copy the url as-is and add `.rss` on the end

packages/fern-docs/bundle/src/components/RSSFeedButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+    <CopyToClipboardButton content={getRssUrl}>
+      {(onClick) => (
+        <FernButton
+          variant="outlined"
+          onClick={onClick}
+          rounded
+          className="[&_.fern-button-content]:gap-2"
+          rightIcon={<Rss className="text-(color:--green-11) size-3.5" />}
+        >
+          Subscribe via <span className="text-(color:--green-11)">RSS</span>
```
chdeskur, line 27: thoughts on using their accent color instead of green?

packages/fern-docs/bundle/src/components/RSSFeedButton.tsx
```diff
@@ -0,0 +1,32 @@
...
+  };
+
+  return (
+    <CopyToClipboardButton content={getRssUrl}>
+      {(onClick) => (
+        <FernButton
+          variant="outlined"
+          onClick={onClick}
+          rounded
+          className="[&_.fern-button-content]:gap-2"
```
chdeskur, line 24: we're aiming to give these net-new features an easily selectable classname, so they can be hidden with CSS if people want to opt out

packages/fern-docs/bundle/src/components/PageHeader.tsx
ariel-serruya: It actually doesn't look half bad on mobile but is a little tight on some of the smaller phones (e.x. iPhone SE). If we think its unnecessary (not sure how common RSS usage is on mobile), I can hide it in the same way we hide `PageActionsDropdown`. Just let me know! 

Screenshot 2025-07-03 at 12 28 29 AM

packages/fern-docs/bundle/src/components/RSSFeedButton.tsx
ariel-serruya: Good call

packages/fern-docs/bundle/src/components/RSSFeedButton.tsx
ariel-serruya: In that case, do we want to hide the RSS button when they enter a specific changelog entry? At that point, the url changes to `http://localhost:3000/learn/docs/getting-started/changelog/2025/6/5` and the .rss suffix is no longer valid. If the `2025/6/5` string is a consistent format across companies, I'm happy to split it out.

packages/fern-docs/bundle/src/components/PageHeader.tsx
ariel-serruya: Removed it on mobile for the time being!

packages/fern-docs/bundle/src/components/RSSFeedButton.tsx
ariel-serruya: Removed it from the entry page for now!

chdeskur: let's leave off entry pages - good call

chdeskur (approved):

PR #3069, opened 2025-07-02
chore: check redirect before pruning

chdeskur:
check auth redirect before pruning tree 

https://github.com/user-attachments/assets/e3e00138-7153-42b1-8399-f3ea43747b2b


mattblank11 (approved):
🚀

PR #3080, opened 2025-07-03
feat: check github repo permissions for visual editor

ariel-langer:

When i was modifying a docs repo in visual editor, i realized that my github user doesn't have write permission to the visual-editor-test-repo which is why i was failing to write to this repo; we should probably handle this failure more elegantly in the future so customers of Fern can also triage the same issue.

NOTE: also FWIW we don't expect to see this case come up much because if a user doesn't have permissions for a repo i dont believe it shows up in this drop down
 
image

Better UX for future users of the visual editor.

looks like this for now:

Screenshot 2025-07-03 at 12 22 24 PM
Screenshot 2025-07-03 at 12 22 19 PM
Screenshot 2025-07-03 at 4 44 36 PM


sbawabe (approved):
lgtm!!

PR #3106, opened 2025-07-08
feat(ve): editable docs loader

sbawabe:
Fixes FER-


sbawabe:
closing out in favor of simpler solution for now, and having later work on supabase fill this in!

PR #3112, opened 2025-07-09
feat(askfern): Define initial `fai-sdk` definition.

eyw520:
This PR creates a new `fern/apis/fai` folder containing an initial `fern definition` for our `fai` microservice.

The `fai-sdk` package is generated into the `packages/fai-sdk` folder.


fern/apis/fai/generators.yml
```diff
@@ -0,0 +1,19 @@
...
+
+groups:
+  local:
+    generators:
+      - name: fernapi/fern-typescript-node-sdk
+        version: 0.51.0
+        output:
+          location: local-file-system
+          path: ../../../packages/fai-sdk/src
+        config:
```
dsinghvi, line 11: lets disable all configs and use the generator as is for now

dubwub (approved):

PR #3113, opened 2025-07-09
feat(ve): sync MDX data from client MdxStateContext to server EditableDocsLoader

stephencme:

Add unidirectional syncing from MdxStateContext -> EditableDocsLoader and clean up some naming conventions. The `setMdxFile` server action (wraps `loader.setMdxFile`) is fired from the context and then calls the loader.

Also adds UX to disallow commits until the state is synced with the server. I think it feels ok as-is, but there may be more sophisticated solutions including:
- Sync state "spinner", Notion-style
- Allow the user to press the commit button immediately, wait for latest sync behind the scenes on the server side

Some other related notes:
- `createEditableDocsLoader` is called in a few different places, with an existing TODO for pulling in the `host`. If we're ok with the way I've approached this in `actions/setMdxFile.ts` then I can fix those in this PR
- We currently key MDX files by filename, but I expect some customers might have duplicate filenames in different parts of their docs. We need a way to disambiguate these files (by path, perhaps?)

Set up a structure for persisting the user's edits, most likely in Supabase. Eventually, there will be more types of data (beyond MDX) to sync from the client -> server using the same structure. The context, loader, and action can be progressively updates as these requirements arise.

To start on Supabase, the work is roughly:
- Merge FDR and Supabase data for the initial SSR of the VE page
- Persist incoming data from the `setMdxFile` action call to Supabase
- Consider whether to move control of the logic in `DashboardApiClient` to the server

Manually on my local machine.


packages/fern-dashboard/src/providers/MdxStateContext.tsx
```diff
@@ -6,51 +6,81 @@ import {
...
+
+  // Stablilize updateDependencies identity to prevent unnecessary re-renders
   const updateDependencies = useCallback(
-    (filePath: FilePath, state: MdxDependencies) => {
+    (filename: Filename, state: MdxDependencies) => {
       setMdxDepsStore((prev) => ({
         ...prev,
-        [filePath]: {
-          html: state.html ?? prev[filePath]?.html,
+        [filename]: {
```
sbawabe, line 73: appreciate you keeping naming consistency with `filename`!! 🚀

sbawabe (approved):
looks great to me! appreciate you adding a clear pr comment here as well 💪

PR #3138, opened 2025-07-11
Posthog tracking events 

ariel-serruya:
I added a new tracking function in our components folder that lets us post custom Posthog events based on user clicks. We can now track `CopyToClipboard`, `Step`, and `SidebarLink` events. 

[This](https://linear.app/buildwithfern/issue/FER-5520/payabli-posthog-tracking-for-copy-page) Payabli suggestion/linear task was the initial impetus. 

https://github.com/user-attachments/assets/9b7cf7e0-c4c0-466c-af15-22267f74711a


packages/fern-docs/components/src/analytics/track.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Track an event.
+ *
```
chdeskur, line 3: weird because i think this code already exists in the codebase, but it should live at this file path so idk why a diff is showing up. can you double check?

PR #3141, opened 2025-07-11
chore: comment out fa-sdk build script

sahil485:
- comment out fa-sdk build script


dsinghvi (approved):
@eyw520 lets remove this temporarily and then we can come back and fix

PR #3143, opened 2025-07-11
feat(ve): more comprehensive preloading logic

sbawabe:
Fixes FER-


dsinghvi (approved):

packages/fern-dashboard/src/app/api/get-docs-github-source/handler.ts
```diff
@@ -21,74 +23,84 @@ export default async function getDocsGithubSourceHandler({
...
-    url: decodeURIComponent(url),
-    token,
-  });
-  if (!docsUrlMetadata.ok) {
-    // the docs url is user-supplied (parsed from the page url) so it's ok if it
-    // doesn't exist
-    if (docsUrlMetadata.error.error === "DomainNotRegisteredError") {
-      return EMPTY_RESPONSE;
-    }
+  // Cache the result for 5 minutes to avoid repeated API calls
```
stephencme, line 26: In practice how often do we expect docs url metadata to change? Do we need to be worried about stale cache?

stephencme (approved):
Did not test locally but code looks reasonable to me. Makes sense to use experimental Next PPR on layout and more aggressively preload data.

@sbawabe curious if you were able to benchmark the actual perceived load time improvements due to this PR?

packages/fern-dashboard/src/app/api/get-docs-github-source/handler.ts
sbawabe: i would say pretty unoften -- i left this as 5 min cache just to dedupe API calls, the big metadata piece that would change would be if someone changes the source repo of the domain while someone else is also in VE for that domain. feels unlikely to me i'd say?

but definitely an edge case we'd want to build protection for later on!

sbawabe:
great q -- i have not. one thing i saw in vercel is the ability to have [speed insights](https://vercel.com/buildwithfern/fern-dashboard/speed-insights). maybe we turn that on for dashboard?

PR #3159, opened 2025-07-14
feat(dash): enable speed insights

sbawabe:
Fixes FER-


stephencme (approved):
+1

PR #3178, opened 2025-07-14
feat(ve): ability to render fern components

stephencme:
Fixes FER-

- adds rendering support for basic fern components into tiptap
- tiptap components do not appear to be stateful, so any stateful components or components that didn't seem to render well were excluded
   - example: tabs components would show nothing, then show all tabs' content at once upon click. clear mismatch in behavior


packages/fern-dashboard/src/app/[orgName]/(visual-editor)/editor/[docsUrl]/[branch]/[...slug]/PageContents.tsx
```diff
@@ -29,7 +31,21 @@ export default function PageContents({
 
   const { updateDependencies, changedMdxFiles, syncChanges } = useMdxState();
 
-  // Set up initial mdx dependencies
+  const { originalElements, setOriginalElements } = useOriginalElements();
+
+  useEffect(() => {
+    // Only bundle if elements don't already have code (prevents infinite loop)
+    const needsBundling = Object.values(originalElements).some(
```
Comment (stephencme) [inline packages/fern-dashboard/src/app/[orgName]/(visual-editor)/editor/[docsUrl]/[branch]/[...slug]/PageContents.tsx:38]: Alternative to this is to do a similar check in bundleOriginalElements and and ensure the identify of originalElements is stable if bundling is not needed. This has the added benefit of being stable if we want to use originalElements as a dependency in other places.

I'm good with this approach though, I think it's clear! A comment noting this alternate approach might be useful if we want to refactor in the future.

packages/fern-dashboard/src/app/[orgName]/(visual-editor)/editor/[docsUrl]/[branch]/[...slug]/bundleOriginalElements.tsx
```diff
@@ -0,0 +1,34 @@
+"use server";
+
+import { bundleMDX as internalBundleMDX } from "mdx-bundler";
+
+import { OriginalElement, OriginalElements } from "@fern-docs/mdx";
+
+export async function bundleMDX(source: string) {
+  const { code } = await internalBundleMDX({
+    source: source,
+    // globals: {
```
Comment (stephencme) [inline packages/fern-dashboard/src/app/[orgName]/(visual-editor)/editor/[docsUrl]/[branch]/[...slug]/bundleOriginalElements.tsx:10]: Sorry this is sloppiness left from the state I passed this to you in @sbawabe but can we remove this commented block?

packages/fern-dashboard/src/app/[orgName]/(visual-editor)/editor/[docsUrl]/[branch]/[...slug]/page.tsx
```diff
@@ -105,17 +106,19 @@ export default async function Page({
           productIsDefault={foundNode.isCurrentProductDefault}
         />
         {foundNode.node.type !== "page" ? (
-          <CustomElementNodeViewClient>
+          <UnsupportedContent>
```
Comment (stephencme) [inline packages/fern-dashboard/src/app/[orgName]/(visual-editor)/editor/[docsUrl]/[branch]/[...slug]/page.tsx:109]: Given the renaming here (and given we're now using it outside of the Tiptap extension for an error boundary) I think it makes more sense for this component to live one level up in src/components/editor

packages/fern-dashboard/tsconfig.json
stephencme, line 1: Assuming that target was removed to get the new components building properly. Given the complexity of our tsconfigs I'd prefer to be explicit about which edition of JS we're targeting. I believe the default target may be `ES5`.

Also nit to remove the commended out compilerOptions (unless there's something you wanted to communicate by including it).

stephencme (commented):
Realized this was originally my PR so I can't approve 😂

Left some notes. Mostly very tiny nits on code quality and organization, no blockers. @sbawabe lmk your thoughts and if you want to finish up this PR yourself or just +1 it for me and I'll get it merged in.

packages/fern-dashboard/tsconfig.json
sbawabe: chatted offline: cleaned up the tsconfig, but the targets were removed so that this tsconfig would match the one in the `fern-docs/bundle`!

sbawabe (approved):
slayyyyyyy

PR #3212, opened 2025-07-16
feat(ve): add placeholder text and better menu open functionality

sbawabe:
Fixes FER-

- adjusts functionality of menu so that it is opened on typing `/` instead of always
- adds placeholder


stephencme (approved):
+1, this is slick! Flagging that any blocks with strings that starts with a `/` character will always show the menu when focused – there's probably a UX that feels better here, but don't think it's worth the complexity atm.

For the future (and related to above): would be cool if typing after the `/` filtered the list of block type options instead, Notion-style.

sbawabe:
### Merge activity

* **Jul 16, 6:24 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/fern-api/fern-platform/3212).

PR #3213, opened 2025-07-16
feat(askfern): Define initial `analytics/insights/{domain}` API layer.

eyw520:
This PR defines the `GET /analytics/insights/{domain}` endpoint, which will return a series of insights on `Ask Fern` queries between a given date range.


edisonmoy (approved):
lgtm

PR #3216, opened 2025-07-16
fix(ve): remove hyperlink button

sbawabe:
Fixes FER-

- comments out hyperlink from floating menus since this is not yet implemented


stephencme (approved):
+1

PR #3236, opened 2025-07-18
fix(ve): fix redirects

ariel-langer:

Discovered that when the feature flag for VE is enabled by orgs it introduces a bug where if you are on `http://localhost:3000/ariel/docs/ariel.docs.buildwithfern.com` and then you toggle to a new org that doesn't have VE enabled, we try to redirect someone to `http://localhost:3000/[NEWORG]/docs/[NEW_ORG_DOCS]` which then re-directs someone back to `/` because the feature flag is disabled for this new org which means its impossible to view any of the settings for that org today because you'll consistently get redirected

The solution for now is to just to redirect to `/orgName/members` instead.

Switching to feature flags by org name revealed a bug that didn't exist before

Tested locally


ariel-langer (commented):
better fix might also just be to catch all invalid paths like `/orgName/adsklajslkdfkalsdkl` and instead of 404-ing we always re-direct to `/orgName/members` but this is the quick win for now

dsinghvi (approved):

PR #3240, opened 2025-07-18
chore(docs): filter authed tabs

chdeskur:
do not show tabs as discoverable unless edge flag is enabled


chdeskur:
before: 
Screenshot 2025-07-18 at 3 24 45 PM

after:
Screenshot 2025-07-18 at 3 25 05 PM

PR #3277, opened 2025-07-23
chore(docs): ensure config passed for endpoint id locator

chdeskur:
ensure the config is accessible in the endpointId function (not just possibly cached)


packages/commons/docs-loader/src/readonly-docs-loader.ts
```diff
@@ -438,54 +438,53 @@ const getAllApisForDomain = async (
...
-        type: "endpoint",
-        endpointId,
-      }
-    );
+const getEndpointById = async (
+  domain: string,
+  apiDefinitionId: string,
+  endpointId: EndpointId,
+  cacheConfig: Required<CacheConfig>
+): Promise<{
```
dsinghvi, line 446: can we use an object here with properties instead of positional params that are easy to mix up

PR #3278, opened 2025-07-23
feat(askfern): Support custom context indexing. 

eyw520:
This PR enables the `/index/{domain}` endpoint, which allows Fern to serve as a CMS for context that should be surfaced in `Ask Fern`, but may not necessarily want to be included explicitly in the developer documentation. Changes include:

- Defining the initial `contexts` table schema.
- Defining the initial `/index/{domain}` endpoint.

This PR is a DRAFT and should not be merged.


fern/apis/fai/definition/index.yml
```diff
@@ -25,11 +25,36 @@ service:
...
+              type: list<string>
+              docs: The context of the document, as a list of strings, that will be indexed.
             content:
               type: string
               docs: The content of the document that will be returned in the tool response
       errors:
         - commons.BadRequestError
         - commons.InternalError
+
+    updateDocument:
```
dsinghvi, line 37: you're going to need a get! but will +1

dsinghvi (approved):

PR #3294, opened 2025-07-28
chore: update filtering for authed products and tabs

chdeskur:
explicitly check for the `everyone` viewer


dsinghvi (changes requested):
can you explain what this PR is trying to enable, before/after (and ideally some tests)

dsinghvi (approved):

PR #3298, opened 2025-07-28
chore: dont cache errors

dsinghvi:
Dont cache errors.


stephencme (approved):
Makes sense conceptually to me! Let's try this.

PR #3337, opened 2025-08-05
fix: Reimplement `fern_token` auth for `/chat`.

eyw520:
This PR reimplements auth for the `/chat` endpoint and enables both header and cookie auth, giving precedence to the header value (if it exists). We use the auth construct returned:

```
authState {
   authed: false,
   ok: true,
   authorizationUrl: undefined,
   partner: undefined
 }
```

And only return unauthorized if `authState.authed AND !authState.ok`


packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/chat/route.ts
```diff
@@ -37,6 +38,13 @@ export async function POST(req: NextRequest) {
   const createdAt = new Date();
   const host = req.nextUrl.host;
   const domain = getDocsDomainEdge(req);
+
+  const { getAuthState } = await createGetAuthStateEdge(req);
+  const authState = await getAuthState(req.nextUrl.pathname);
+  if (authState.authed && !authState.ok) {
```
Comment (chdeskur) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/chat/route.ts:44]: would double-check the expected data shape of an auth failure. 

things to consider:
- check `!authState.ok` separately?
- check `authState.authed` separately + confirm user roles restrict

Comment (eyw520) [reply packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/chat/route.ts]: Just checked & updated. Data shape of auth failure is just `!authState.ok`. Updated accordingly.

chdeskur (approved):
lgtm

PR #3347, opened 2025-08-06
fix(ve): fix dev mode content for new pages

sbawabe:
Fixes FER-

previously i was making an assumption that we knew the filename because there was only ever one loaded at a time. with the addition of new pages, this yielded new pages to not show the proper content in dev mode.

the fix: add in a quick provider that stores the name of the current filename, so that we can set it and refer to it easily to ensure we always populate with the correct file content!

bug fix

tested locally


stephencme (approved):
Lgtm

PR #3350, opened 2025-08-06
chore: add environment-specific explorer state

chdeskur:
proposal for an extension to our playground auto-population logic. 

currently, only one initial state is tracked. 

this would add the option to set environment-specific values. 
note: 
- env_state values are favored over initial_state values
- initial_state values are used as a fallback (if present) if the environment doesn't match
- the underlying shape of both schemas is the same, so we can reuse logic as much as possible


sbawabe (approved):
lgtm! calling out that it's a bit weird UI (not new to this pr) that if i input a custom token, it's odd that it gets saved in the request but not saved in the auth panel

image

PR #3368, opened 2025-08-11
chore: upload with dynamic ir with api sources

chdeskur:
separates dynamic-ir + sources as a conceptual input, but combines them in the return 

dynamic-ir will be stored alongside the api


servers/fdr/src/services/s3/S3Service.ts
```diff
@@ -311,27 +314,48 @@ export class S3ServiceImpl implements S3Service {
...
+          presignedUrl: url,
+          key,
+        };
+      }
     }
+
+    if (dynamicIr) {
+      for (const { language } of dynamicIr) {
+        const { url, key } =
+          await this.createPresignedApiDefinitionSourceUploadUrlWithClient({
```
dsinghvi, line 346: this feels weird to use for dynamic IR, can we create a function for dynamic ir specifically?

chdeskur: on it

PR #3373, opened 2025-08-11
feat(ve): authed docs in ve!

sbawabe:
Fixes FER-6024

- adds in ability to view authed docs in VE!

- needed for beta

- go to an authed docs site in editor (can use `elevenlabs-internal` and give fake branch name as workaround)
- i also tested with `fern-internal` docs to ensure that auth was still in place as expected for readonly


sbawabe:
* **#3373** <a href="https://app.graphite.dev/github/pr/fern-api/fern-platform/3373?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/fern-api/fern-platform/3373?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `app`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

packages/commons/docs-loader/src/readonly-docs-loader.ts
```diff
@@ -1017,19 +1018,28 @@ export const createCachedDocsLoader = async (
...
-      host,
-      domain,
-      fern_token,
-      await authConfig,
-      await metadata
-    );
-    return await getAuthState(pathname);
-  });
+  const getAuthState = skipAuth
+    ? cache(async (_pathname?: string) => ({
```
stephencme, line 1027: cache(...) is unnecessary here when skipAuth is true since we're just returning an object from this function. My preferences would be to remove the cache call for clarity.

packages/commons/docs-loader/src/editable-docs-loader.ts
```diff
@@ -158,8 +158,13 @@ class EditableDocsLoaderImpl implements EditableDocsLoader {
     return this.readOnlyDocsLoader.getLayout();
   }
 
-  async getAuthState(pathname?: string): Promise<AuthState> {
-    return this.readOnlyDocsLoader.getAuthState(pathname);
+  async getAuthState(_pathname?: string): Promise<AuthState> {
```
stephencme, line 161: This seems redundant – since we've already implemented this skipAuth override in readonly-docs-loader, I suggest we call `return this.readOnlyDocsLoader.getAuthState(pathname);` directly!

stephencme (changes requested):
[q] any security concerns with the skipAuth approach? Is the idea here that "if you have access to the Fern org, you should have access to edit the docs"?

Flagging for myself – we will need to consider the intersection of docs site permissions and GitHub repo permissions once we switch to git backend.

PR #3389, opened 2025-08-13
feat(askfern): `/reindex` triggers sync; `/chat`uses query index.

eyw520:
This PR implements the following:
- Couple `/reindex` with `/sync` to propagate docs index updates to the query index.
- Update `/chat` to consume the query index.


packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/reindex/turbopuffer/route.ts
```diff
@@ -93,6 +98,15 @@ export async function GET(req: NextRequest): Promise<NextResponse> {
       },
       deleteExisting,
     });
+    const faiClient = new FernFaiClient({
+      baseUrl: getFaiOrigin(),
+      token: () => "",
```
Comment (dsinghvi) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/reindex/turbopuffer/route.ts:103]: is this right?

Comment (dsinghvi) [reply packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/reindex/turbopuffer/route.ts:103]: I think it would make sense to init it with `admin_fernToken()` which we do for fdr calls (and in a follow up you can add token verification via Venus in FAI)

packages/fern-docs/search-server/ask-fern/src/turbopuffer/utils/get-turbopuffer-namespace.ts
```diff
@@ -4,7 +4,17 @@ import { withoutStaging } from "@fern-api/docs-utils";
...
+  indexName: string
+): string {
+  return `${withoutStaging(domain)}_${indexName}`;
+}
+
+export function getDocsIndexName(
   embeddingModel: EmbeddingModel<string>
 ): string {
-  return `${withoutStaging(domain)}_${embeddingModel.modelId}_v3`;
+  return `${embeddingModel.modelId}_v3`;
```
dsinghvi, line 15: is removing the domain intentional here?

packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/reindex/turbopuffer/route.ts
Comment (eyw520) [reply packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/search/v2/reindex/turbopuffer/route.ts:103]: This is correct at the moment since we're not using any fernToken auth on `fai`. Will plug with admin_fernToken() until we add auth.

PR #3399, opened 2025-08-14
chore(fdr): Add `python` fdr sdk generation group.

eyw520:
This PR adds a `python-sdk` generation group to the FDR generators.yml config.


fern/apis/fdr/generators.yml
```diff
@@ -46,6 +46,21 @@ groups:
             devDependencies: 
               "typescript": "~5.7.2"
 
+  fdr-cjs-sdk-py:
```
fern-support, line 49: would just call this python-idk

fern/apis/fdr/generators.yml
```diff
@@ -46,6 +46,21 @@ groups:
...
+    generators:
+      - name: fernapi/fern-python-sdk
+        version: 1.4.0
```
fern-support, line 52: can you use the latest version? (or does it not build)

eyw520: Will upgrade

PR #3400, opened 2025-08-14
feat(ve): DO NOT MERGE delete pages feature PoC

stephencme:
Fixes FER-


stephencme:
Vibe-coded this as an exploration. Shipped #3438 as a first step instead, which is much better implemented.

PR #3420, opened 2025-08-17
chore: remove dynamic layout

chdeskur:
since the `layout.tsx` and `page.tsx` are the same components, i propose removing the layout for the following reasons:
- there is nothing about the layout that is optimized for persisting content across shared pages (since we need to recalculate the slug + metadata for each page)
- this could lead to conflicting metadata, which causes `generateMetadata` to function improperly
- it matches our approach for static pages


chdeskur:
next.js docs on layout lead me to believe our usage of `layout.tsx` is incorrect: https://nextjs.org/docs/app/getting-started/layouts-and-pages#nesting-layouts

eyw520 (approved):

PR #3455, opened 2025-08-20
fix tailwind imports in visual editor

coltondotio:

Prior to this PR, devs were experiencing [this issue](https://github.com/vercel/next.js/issues/50388) when developing locally and building the visual editor.

I asked Claude to analyze the tailwind imports, and it showed:
- the visual editor index.css file had `@import "@fern-docs/components/styles";` and `@import "@fern-docs/search-ui/styles";`
- `@fern-docs/search-ui/styles` had `@source "../../components";`, which pointed to `@fern-docs/components`

So, perhaps having both an `@import` and a `@source` to the same CSS file causes tailwind to freak out? I'm not totally sure. But, this fixes the issue!

I'm not sure why this specifically caused the issue, but this PR seems to fix it.


dsinghvi (approved):

stephencme (approved):
Lgtm! I suspect the underlying issue here has to do with @source – this directive registers the specified path with the Tailwind compiler then the compiler attempts to generate the necessary CSS based on the classNames it detects:
https://tailwindcss.com/docs/detecting-classes-in-source-files#explicitly-registering-sources

Given how huge our project is, I wouldn't be surprised if the number of files it's scanning is simply too large to work reliably.

Afaik we don't need any of the components from search-ui in VE, so removing this seems fine to me cc @sbawabe if you know otherwise.

sbawabe (dismissed):
unfortunately this borks searchbar styling;(

before:
Screenshot 2025-08-20 at 9 10 27 PM

after:
Screenshot 2025-08-20 at 9 09 41 PM

stephencme (changes requested):
Un +1'd based on feedback from @sbawabe about the dependency on search-ui styles 👀

stephencme (approved):
+1

PR #3468, opened 2025-08-21
chore(ve): tweak ui to keep ve above fold

sbawabe:
Fixes FER-


stephencme (approved):
Tested preview link and lgtm

PR #3476, opened 2025-08-22
[wip] feat(visual editor): use organization to determine repo auth instead of user

coltondotio:

GitHub repository authorization is now determined based on the requesting Auth0 organization, instead of the requesting user's linked GitHub account.

Currently, the only way to sign into the Fern dashboard is with a GitHub account. We want to expand this so that users can log in with their org's SSO, or Google, etc.

To facilitate this expansion of login methods, this PR removes the requirement that a user's account is associated with a GitHub account.
- Specifically, we were leveraging this GitHub association to determine if a user had write permissions to the GitHub repositories they are editing. Now, instead, we leverage the `organization` field of the `./fern/fern.config.json` file to determine which Auth0 organization is allowed to write to the repository.

Still testing it!


packages/fern-dashboard/src/app/[orgName]/(homepage)/docs/[docsUrl]/page.tsx
```diff
@@ -60,9 +60,13 @@ export default async function Page(props: {
...
-    repoExists: false,
-    hasWriteAccess: false,
-    hasFernBotInstalled: false,
+    validationResult: {
+      ok: false,
+      error: {
+        type: "UNEXPECTED_ERROR",
+        message: "",
+      },
+    },
```
Comment (sbawabe) [inline packages/fern-dashboard/src/app/[orgName]/(homepage)/docs/[docsUrl]/page.tsx:69]: i do really like the use of result type for validation here -- it gives a clear "are you allowed yes/no" / "if no, why"! 🚀

packages/fern-dashboard/src/app/[orgName]/(homepage)/docs/[docsUrl]/page.tsx
```diff
@@ -149,7 +149,9 @@ export default async function Page(props: {
             </p>
           </div>
 
-          {!githubAuthState.hasFernBotInstalled ? (
+          {!githubAuthState.validationResult.ok &&
```
Comment (sbawabe) [inline packages/fern-dashboard/src/app/[orgName]/(homepage)/docs/[docsUrl]/page.tsx:152]: now that we have a new case of "unknown error", i want to make sure we don't do success as `else` anymore. can we swap this to:

```
if validation.ok:
    <GoToEditor>
else if:
    // existing stuff
else: // new unknown error case
    // maybe warning note with "we were unable to validate access to this repo"

File: packages/fern-dashboard/src/app/services/dal/github/validators.ts [hunk 3]
```diff
@@ -103,39 +89,49 @@ export const validateGithubRepoAccess = async (
...
-      "FERN_BOT_NOT_INSTALLED"
-    );
+  const validation = await validateGithubRepoAccess(orgName, identifier);
+
+  if (validation.ok) return;
+
+  const { error } = validation;
+  let message: string;
+
+  switch (error.type) {
```
stephencme, line 107: We already define a similar error type to message mapping in `middleware.ts`. Are there cases where we'd want this mapping to deviate between implementations?

If not, we can just pass `validation.error.message` through.

packages/fern-dashboard/src/components/editor/CommitButton.tsx
```diff
@@ -350,6 +352,10 @@ export function CommitButton() {
       ErrorNoBranchToast();
       return;
     }
+    if (!orgName) {
+      ErrorNoBranchToast();
+      return;
```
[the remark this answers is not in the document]
stephencme: Seems like a fair point!

PR #3482, opened 2025-08-22
[wip] 2025 08 22 stephen complex yml

stephencme:
Fixes FER-


stephencme:
Continued this work on #3523 after realizing that my original approach here could be improved.

PR #3569, opened 2025-09-03
Sarah/images without original elements

sbawabe:
Fixes FER-


packages/fern-dashboard/src/components/editor/extension-custom-element/create-custom-element-node.ts
```diff
@@ -0,0 +1,17 @@
+import { createHash } from "crypto";
+
+export function createCustomElementNode(tagName: string, mdxContent: string) {
+  return {
+    type: "custom-element-v2",
+    // These data attributes help the client to handle the custom element
+    attrs: {
+      "fve-data-hash": createHash("sha256")
+        .update(Math.random().toString())
```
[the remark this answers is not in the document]
sbawabe: added

packages/fern-dashboard/src/components/editor/TiptapEditor.tsx
```diff
@@ -78,7 +83,11 @@ export default function TiptapEditor({
   return (
     <EditorProvider
       autofocus={autofocus}
-      extensions={extensions}
+      extensions={[
```
stephencme, line 86: [nit] prefer that these extensions be loaded in where the extensions variable is assigned so that the configuration can be read in one place

packages/fern-docs/mdx/src/convert.ts
stephencme, line 1: Flagging that without originalElements, we’re going to lose the diffing optimizations I worked on and which were a major design constraint of the original system. I anticipate we'll see a regression in diffs and will need to work on a solution for this again.

Did y'all have ideas for different approach to the problem of preserving the original formatting of a node? The underlying problem with the diffs is that two markup strings could be semantically the same, but formatted differently (and we need a way to preserve the original formatting).

Instead of removing the originalElements code from this file, I'd suggest keeping it (but making it optional) so that we can leverage it when we loop back to fix the diffs regressions in the future.

packages/fern-dashboard/src/components/editor/extension-custom-element/custom-element.ts
```diff
@@ -81,93 +85,4 @@ export const CustomElement = Node.create<CustomElementOptions>({
       0,
     ];
   },
-
-  addProseMirrorPlugins() {
```
stephencme, line 85: Discussed in Slack – we want to keep (or modify) the behavior here rather than remove it entirely.

PR #3580, opened 2025-09-04
feat(ve): support multiple fern projects in one git repo

coltondotio:
feat(ve): support multiple fern projects in one git repo

many Fern users have a repo containing multiple fern projects, like:
```
foo/
  fern/
bar/
  fern/
```

this was previously incompatible with the dashboard / visual editor.

now, we use the current `site` (docsUrl) to determine which specific fern project should be used when fetching `docs.yml` or `fern.config.json`.

locally, i have tested many cases:
- [invalid] a repository that i don't have access to
- [valid] a repository with 2 fern projects, one of which corresponds to the site i was editing
- [invalid] a repository with 2 fern projects, neither of which correspond to the site i was editing
- [invalid] a repository with 2 fern projects that are each configured with the site i was editing

there is a custom message displayed for each of these cases (like the attached picture)

image


packages/fern-dashboard/src/app/[orgName]/(homepage)/docs/[docsUrl]/page.tsx
```diff
@@ -192,11 +195,12 @@ interface ValidationErrorHandlerProps {
   error: GithubRepoValidationError;
   githubUrl?: string;
   orgName: Auth0OrgName;
+  site: string;
 }
-
 function ValidationErrorHandler({
   error,
   orgName,
+  site,
```
Comment (sbawabe) [inline packages/fern-dashboard/src/app/[orgName]/(homepage)/docs/[docsUrl]/page.tsx:203]: [nit] this `ValidationErrorHandler` is becoming fairly meaty -- might be time for us to move this into its own file so that `page.tsx` is easier to parse at a glance!

packages/fern-dashboard/src/app/services/dal/github/middleware.ts
```diff
@@ -14,6 +14,7 @@ import { validateGithubRepoAccess } from "./validators";
 export interface ParsedRepoData {
   owner: string;
   repo: string;
+  site: string;
```
sbawabe, line 17: is `site` the domain associated with the repo? might be nice to keep the terminology consistent with what's in `docs.yml` if so

sbawabe (commented):
looking sweet!! left a couple non-blocking nits, but only issue i noticed is with our own fern repo:

Screenshot 2025-09-04 at 4 56 33 PM

looks like we should have it configured in docs.yml fine (screenshot below), but im seeing the above error. might be a subpath domain issue?

Screenshot 2025-09-04 at 4 56 58 PM

packages/fern-dashboard/src/app/services/dal/github/middleware.ts
coltondotio: no, this is sort of me piggybacking on the existing GithubAuth data stuff -- the idea is that we can't do github repo auth until we know the site that the user is trying to do work on behalf of

so, instead of the client sending `{ owner, repo }`, the client now sends `{ owner, repo, site }`, where `site` is the `docsUrl` they are editing

packages/fern-dashboard/src/app/services/github/github-loader.ts
```diff
@@ -78,62 +91,332 @@ export class GitHubLoader implements GitLoader {
...
+      }
+
+      throw error;
+    }
+  }
+  /**
+   * Finds a Fern project by site URL using tree searching methodology.
+   * Returns the paths to both docs.yml and fern.config.json for the matching project.
+   */
+  async getFernProjectBySite(
```
stephencme, line 156: Love this approach 💪

stephencme (approved):
[nit] warning icon is comically tiny:

image

Otherwise lgtm!

PR #3582, opened 2025-09-04
fix(ve): full support creating pages in tabs navigation

stephencme:

User should now be able to create pages inside of tab navigation and have those pages render properly.

image

This has been broken for a while and is now a simpler solve with the new #3523 work.

Manually across a few different sites in the dashboard.


stephencme:
> lgtm! only comment is would love this copy to be changed, and perhaps have this tooltip be on the bottom side so that it doesn't block other header tabs
> 
> Screenshot 2025-09-04 at 5 07 21 PM

We decided to just hide the button in these cases instead!

PR #3612, opened 2025-09-09
feat(generator-cli): Ability to add custom sections with template variables such as `version` and `packageName`

jsklan:
First step in fixing [FER-6380](https://linear.app/buildwithfern/issue/FER-6380/)

- Added property to ReadmeConfig for CustomSections in generator cli fern definition
- Updated ReadmeGenerator Logic to handle this case
- Also fixed a bug in how the ReadmeGenerator merges blocks. Existing behavior would have newly added blocks appended to the end of the readme, no matter what. This would mean if a user tried to turn on a new feature on an already generated readme, the new feature section would appear below the contributing section. Adjusted merge ordering logic and added a test case to demonstrate the behavior.

Users want to be able to have custom sections that include a reference to the latest version of the sdk that updates along with the rest of the README. After this PR merges, changes will be made to the fern cli and generators to allow users to input markdown into the `readme` section of `generators.yml`

Added fixture snapshot tests to generator-cli


chdeskur (approved):

packages/generator-cli/src/readme/ReadmeGenerator.ts
```diff
@@ -926,3 +988,7 @@ function getMajorVersion(version: string): string {
 function assertNever(x: never): never {
   throw new Error(`unexpected value: ${JSON.stringify(x)}`);
 }
+
+function toScreamingSnakeCase(s: string): string {
```
Swimburger, line 992: If we already import lodash, we could use their built-in utility

PR #3660, opened 2025-09-15
feat(dashboard): register orgName as a PostHog super property

stephencme:

So we can filter events like page views in PostHog by org.

Confirmed that I could send these events from dev env to PostHog successfully.


sbawabe (approved):
slayyyyy

PR #3683, opened 2025-09-18
fix(dashboard): show correct error message when repo is not connected

coltondotio:

show correct error message when repo is not connected

tested locally


stephencme (approved):
+1 but could you please remove the console.log before merging?

PR #3684, opened 2025-09-18
feat(fe): delete any page, NavigationStore refactor

stephencme:

This feature branch adds support for deleting pages from the sidebar UI. previously, it was only possible to delete pages that were created in current session from the client side.

Full list of changes in this PR:

- **Adds ability to delete pages originating from the initial FDR snapshot**
- Adds toast which gives user ability to undo recent deletes
- **Removes PagesStore in favor of unified NavigationStore API**
- Adds migration to preserve user data from the old localstorage data model
- **Adds `forceClientRender` flag to the sidebar so it can be explicitly rendered on the client in Dashboard, while preserving exsting behavior in Docs**
- Introduces fractional indexing for client pages, used to determine order of client pages in sidebar; foundation for future page reordering feature
- Cleans up navigation utils

Bug fixes:

- **Fixes pages without edits getting marked as changed**
- Fixes Dev Mode for client pages
- **Fixes bugs where commits fail due to missing base docs.yml, reduces unnecessary GitHub API calls by improving docs.yml loading strategy**

Outstanding items to discuss:

- [x] This PR should migrate user data from the old NavigationStorage (localStorage) schema, but it hasn't yet been tested with real user data
- [x] ~This PR marks `SidebarPageNode` as a client component: how should I benchmark this to ensure it doesn't impact Fern Docs performance?~

Update: I've done more testing of the migration logic and it seems to be in good shape, but we should do more smoke testing before deploying.

Customers want this delete feature + PagesStore was still hard to use, even after #3523 

Thoroughly on my machine.


package.json
```diff
@@ -12,6 +12,7 @@
     "codegen": "turbo codegen",
     "compile": "turbo compile",
     "dashboard:dev": "turbo dashboard:dev",
+    "dashboard:dev:turbopack": "turbo dashboard:dev:turbopack",
```
sbawabe, line 15: niceee

sbawabe (approved):

PR #3712, opened 2025-09-21
fix: add in use cache for select get requests

abvthecity:
Fixes FER-


coltondotio (commented):
i'm not sure if we can use cacheLife since we aren't on a canary version anymore https://nextjs.org/docs/app/api-reference/functions/cacheLife

PR #3760, opened 2025-09-24
fix: disable auth on settings endpoint

sahil485:
- disable auth on settings endpoint


servers/fai/src/fai/routes/settings.py
```diff
@@ -30,7 +30,6 @@
 async def get_settings(
     domain: str,
     db: AsyncSession = Depends(get_db),
-    _: None = Depends(verify_token),
```
dsinghvi, line 33: can we make sure to cache verify_token in some small TTL cache in a follow up

PR #3779, opened 2025-09-25
fix(ve): fix fdr stale caching issue

sbawabe:
Fixes FER-6595

- add true `uncachedLoadWithUrl` fn so that we can keep docs using cached value and use uncached value within dashboard

- prevent cache staleness when opening new sessions in Fern Editor after having just deployed docs site

- tested locally & in review app with `sarahbawabe` org
- i randomly had FE error at one point -- seems to be related to experimental ppr, that will be removed in separate pr by @coltondotio


sbawabe:
### Merge activity

* **Sep 25, 10:05 PM UTC**: @sbawabe merged this pull request with [Graphite](https://app.graphite.dev/github/pr/fern-api/fern-platform/3779).

PR #3977, opened 2025-10-06
chore: revalidate should include host fallback

chdeskur:
this logic existed in the readonly-docs-loader, but since we cache the api endpoint definition on revalidation (which happens when you register as docs site as well), the readonly-docs-loader codepath wasn't hit.


packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/revalidate/route.ts
```diff
@@ -154,17 +155,21 @@ export async function GET(
...
-                        prunedApi.forEach((value, key) => {
-                            keys[`api:${key}`] = value;
+                        prunedApi.forEach((api, key) => {
+                            
+                            for (const endpointK of Object.keys(api.endpoints)) {
+                                if (api.endpoints[EndpointId(endpointK)]?.environments?.length === 0) {
+                                    console.debug(`${endpointK} has empty environments, adding default URL.`);
+                                    api.endpoints[EndpointId(endpointK)]?.environments?.push({
+                                        id: "Default" as EnvironmentId,
+                                        baseUrl: "https://host.com"
```
Comment (kennyderek) [inline packages/fern-docs/bundle/src/app/[host]/[domain]/api/fern-docs/revalidate/route.ts:167]: could you explain why this is `https://host.com`?

PR #4015, opened 2025-10-07
fix(analytics): Persist endpoint route to Posthog in Playground

jischein:
Make sure we track which endpoint users hit when using the Playground.


dannysheridan:
testing plan: try it in prod!

